PR #1, opened 2022-03-16
added_flip_tile_and_deck_cover_components

derricklee43:
Added initial flip tile and deck cover components

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,42 +1,11 @@
+@import url('https://fonts.googleapis.com/css2?family=Comfortaa:wght@300;700&display=swap');
+
+:root {
+  background-color: #1d1d21;
+}
+
 .App {
-  text-align: center;
-
-  p span {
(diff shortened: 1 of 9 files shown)


jiejasonliu (approved):
LGTM.

One thought, I recall discussing about this and maybe I forgot the solution but in order to support multiple decks, don't we need some way to distinguish them from another? 

For example, we would need a way to tell which deck a user selects in order to decide which one should be "flipped" (active).

derricklee43:
Yeah, we do need to distinguish them. I was thinking of doing this in the parent component. The current parent component is just a dummy. We could do something like this: 
```
decks.map((deck) => 
     <DeckCover 
           deck={deck} 
           isActive={deck === activeDeck} 
           onClick={()=>{setActiveDeck(deck)}}
           ... 
      />
})
```
We would have a more complex onClick handler, but something like that.  What do you think?

PR #3, opened 2022-03-21
created progress bar component and added it to deck cover

derricklee43:


src/components/ProgressBar/ProgressBar.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  .outer-bar {
+    background-color: #ffff;
+    border: 1px solid #25283d;
+    border-radius: 10px;
+    height: 10px;
+    width: 100%;
+
+    .inner-bar {
```
jiejasonliu, line 17: There's a bit of a UI bug here since the `outer-bar` has a white background, it leaves this white-space:
image

A proposed fix is to inherit the border and set a negative margin in `inner-bar` of the same size to offset the border swallowing the space:
```css
.inner-bar {
    border: inherit;
    margin: -1px;    // should be same size as border
    min-width: 10px; // increase a little to offset smaller margin
}
```

Results in:
image

src/components/ProgressBar/ProgressBar.scss
```diff
@@ -0,0 +1,32 @@
...
+      background-color: #25283d;
+      border-radius: inherit;
+      height: 10px;
+      min-width: 5px;
+    }
+  }
+
+  &.labelOnRight {
```
jiejasonliu, line 25: Did we want to rename this to `label-on-right` to stay consistent with our classes?

src/components/ProgressBar/ProgressBar.scss
jiejasonliu, line 17: There's a bit of a UI bug here since the `outer-bar` has a white background, it leaves this white-space:
image

A proposed fix is to inherit the border and set a negative margin in `inner-bar` of the same size to offset the border swallowing the space:
```css
.inner-bar {
    border: inherit;
    margin: -1px;    // should be same size as border
    min-width: 10px; // increase a little to offset smaller margin
}
```

Results in:
image

jiejasonliu: To add to this, I noticed that setting the progress bar to `0%` still has this filled due to the `min-width`.
We could potentially hard code `min-width: 0` in the code if `percent = 0`, but up to your judgment.

jiejasonliu (changes requested):

src/components/ProgressBar/ProgressBar.scss
derricklee43: Yeah, you're right.  I've updated it to be `label-on-right` for this class name and updated the variant interface for it.

src/components/ProgressBar/ProgressBar.scss
derricklee43: Thanks for catching that, haha. I got it changed with your suggestions. 

When I set the min-width to 0, the inner bar still appeared for me and broke through the outer bar. 

I added some logic for ProgressBar to not render the inner bar when the percent is 0. 
```
{percent > 0 && <div className="inner-bar"> ... }
```

jiejasonliu (approved):

PR #4, opened 2022-03-21
Add variants to button

derricklee43:
Also fixed file names


jiejasonliu (approved):
Not sure why your git didn't pick up case changes, but they're updated now! 
Looks good!

PR #5, opened 2022-03-22
added partial class and swapped in variables in other files

derricklee43:
I standardized and named all colors in the mockups. I used those exact names in the partial file to easily map the colors we need.


jiejasonliu (approved):
I really like this change, 👍

PR #7, opened 2022-03-24
Add drop down

derricklee43:
- added dropdown menu
  - label, option, and buttonLabel are typed as React.ReactNode instead of string to allow passing in JSX into component
  - buttonLabel can be used as to display selected option or something else
- changed color variables to be consistent with CSS naming conventions
- added _styles partial for mixins and standard stylings

https://user-images.githubusercontent.com/39753553/159990201-650105b8-0ac5-4c66-b776-ed1747335941.mov


src/components/drop-down/drop-down.scss
```diff
@@ -0,0 +1,69 @@
...
+    border-radius: $standard-radius + 1px; //hides corners of div
+    @include box-shadow;
+
+    .option {
+      width: 100%;
+      box-sizing: border-box;
+      border-left: 1px solid $flashcard-white;
+      border-right: 1px solid $flashcard-white;
+      background-color: $white;
+      overflow: scroll;
```
jiejasonliu, line 48: This might be problematic, this would make the scrollbars always show -- we also probably don't want text to go past the option or label.

Here's what it looks like now:
current-overflow

We could try something with overflow ellipsis (perhaps make a mixin for this) and apply this to this `.option` here as well as `.drop-down-menu > button > label`
```css
overflow: hidden;
white-space: nowrap;
text-overflow: ellipsis;
```

fixed-overflow

jiejasonliu: Another related note, it doesn't seem like the drop-down-menu label is picking up the font? Perhaps we need to add the font-family to it again 😕

jiejasonliu (changes requested):
Love it so far, just one thing that might need potential tweaking

src/components/drop-down/drop-down.scss
derricklee43: haha, yeah that looks awful, I got that fixed to use the ellipsis styling now. Weird that it wasn't showing up on my browser.

jiejasonliu (approved):

jiejasonliu:
LGTM!

PR #9, opened 2022-03-28
Add text area component

jiejasonliu:
Two variants:
- readonly
- not... readonly

Philosophy: 
- consumer should subscribe to `onChange` and save the state on their own
- consumer needs to know what is in the textarea at anytime since there's no "submit/enter" logic

text-area


src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const TextArea = ({
+  lines,
+  initialText,
+  label,
+  placeholder,
+  readonly,
+  resizable,
+  onChange,
+}: TextAreaProps) => {
+  const [value, setValue] = useState(initialText);
```
derricklee43, line 24: Could we move this out of the component and up a level? So have a prop `value` instead `initialText`. With the import and export dialog, we will need to be able to set the text repeatedly

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <div className="c-text-area-wrapper">
+      {label && <label className="c-text-area-label">{label}</label>}
+      <textarea
+        className={`c-text-area ${readonly ? 'readonly' : ''}`}
+        rows={lines}
+        placeholder={placeholder}
```
derricklee43, line 32: That's nice that nothing shows up when the placeholder is undefined. I'll going to simplify that in mine, haha

derricklee43 (changes requested):

src/components/text-area/text-area.tsx
jiejasonliu: Yeah, I didn't think about a way of doing that without using `useRef` but I think this makes sense now.

src/components/text-area/text-area.tsx
jiejasonliu: Yup, as long as it's not a field that makes the component uncontrolled, `undefined` will act as if it just ignored the attribute.

src/components/text-area/text-area.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div className="c-text-area-footer">
+        <LockIcon />
+        <label>readonly</label>
+      </div>
+    );
+  }
+
+  function onChangeHandler(e: ChangeEvent<HTMLTextAreaElement>) {
+    onChange?.(e.target.value);
```
derricklee43, line 50: I just noticed this. That's so dope that you can use the optional chaining operator with function calls. I'm still this too😂

src/components/text-area/text-area.tsx
derricklee43: That makes it easier on our end, haha

derricklee43 (approved):
LGTM!

src/components/text-area/text-area.tsx
jiejasonliu: ahahaha wow, you're stealing everything!!! 😡😂

PR #10, opened 2022-03-28
Improve circleci caching

jiejasonliu:
Subtle tweak but significant when we start writing more tests

- Cache key doesn't depend on branch, just pure checksum on package/lock.json
- Skip `npm ci` altogether if cache matches

Speedup is around 2x for a cached workflow


derricklee43 (approved):
Nice!! I'm glad you know what you are doing with these pipelines😆

PR #12, opened 2022-03-30
Improve search bar component

jiejasonliu:
Breadth of changes here is relatively large, doesn't need to merge immediately, we can iterate on this

Preface that there's a lot of state being thrown around...
Many areas can probably be improved, if you see anything low hanging please lmk

Significant changes:
- Add `onDebouncedChange` to batch changes if we need
- Show dropdown when search matches values in `dropDownData: DropDownOption[]`
- Dropdown selection will _autocomplete_ the search field and notify change events
- Add `useOutsideClick` hook to perform a side effect when the user clicks outside of the `ref` you set
    - (e.g.) hide the dropdown when the user clicks outside of the element set as the ref
    
search-bar-dropdown-30fps


src/components/search-bar/search-bar.scss
```diff
@@ -6,54 +6,98 @@
...
-    height: $height;
+  .c-search-bar-content {
+    display: inline-flex;
+    justify-content: space-between;
+    width: 100%;
 
-    &.has-text {
-      border-right: 1px solid $light-blue;
-      background-color: $blue;
+    .c-search-icon-container {
```
derricklee43, line 21: I like your naming convention for components. I'll start doing that too, and I'll go back and update the other classes as well

src/components/search-bar/search-bar.tsx
```diff
@@ -1,52 +1,130 @@
...
+  const [hasClickedSinceLastChange, setHasClickedSinceLastChange] = useState(false);
+
+  const debouncedChange = useMemo(
+    () => debounce(onDebouncedChange ?? noop, debounceMs),
+    [onDebouncedChange, debounceMs]
+  );
+
+  // hide dropdown after clicking outside the search bar wrapper div
+  const wrapperDivRef = useRef(null);
+  useOutsideClick(wrapperDivRef, () => setHasClickedSinceLastChange(true));
```
derricklee43, line 45: This useOutsideClick is so nice. I think I will add that to the flashcard component if we decide to do an onBlur thing

derricklee43: I didn't realize you made that yourself; that's so dope

derricklee43 (approved):
LGMT!

src/components/search-bar/search-bar.scss
jiejasonliu: Cool! I ran into a conflict in class namespaces and decided to add c- to all classes in our components to avoid that. :P

PR #13, opened 2022-03-30
Add flashcard

derricklee43:
- 3 variants: active, inactive, readonly
- sorry for the off audio in the video
- component uses audio passed in through the CardContent interface
- is set to active onFocus to allow keyboard navigation
  - There is no onBlur to unselect a flashcard (I'm not sure if we will even want that in the future)
- I made app.tsx and app.scss a mess (we can clean it up with the real flashcard parent)
- I plan on going back and adding tests after the demo
- A few of the icons had some weird stuff for changing the colors, so I created icon components for them. 
  - I am not sure if you want all svgs to be standardized, if so we can make them all into icons, or move the logic from the ones I created into their respective components. 

https://user-images.githubusercontent.com/39753553/160858527-36e0c4d1-2a62-418a-b80f-06e47927fa56.mov


src/app.tsx
```diff
@@ -100,8 +162,65 @@ function App() {
...
+    });
+
+    function handleRemoveClick(id: number) {
+      const newCards = [...cards];
+      newCards.splice(id, 1);
+      setCards(newCards);
+    }
+
+    function handleUpClick(id: number) {
+      // Todo handle up click when card is first
```
jiejasonliu, line 205: This actually works already, negative indices in `splice` define the position from the end of the array.

src/app.tsx
```diff
@@ -100,8 +162,65 @@ function App() {
...
+      const newCards = [...cards];
+      newCards[id] = newCards.splice(id - 1, 1, cards[id])[0];
+      setCards(newCards);
+    }
+
+    function handleDownClick(id: number) {
+      //Todo handle down click when card is last
```
jiejasonliu, line 212: Is this something you want to fix before demo/merging this? 

If not, can you track this either by opening an issue or adding it to our trello board?

src/components/flashcard/card-face/card-face.tsx
```diff
@@ -0,0 +1,63 @@
+import './card-face.scss';
+import React from 'react';
+import { ReactComponent as Speaker } from '../../../assets/svg/speaker.svg';
+import { ReactComponent as KebabMenu } from '../../../assets/svg/kebab-menu.svg';
```
jiejasonliu, line 4: I think for svgs/icons, we should mark it as ...Icon, so (SpeakerIcon) and (KebabMenuIcon) so we don't mistaken it for a functional component.

src/app.tsx
```diff
@@ -100,8 +162,65 @@ function App() {
...
+          onFocus={() => setActiveCard(card.id)}
+          onCardChange={(c) => handleCardChange(c, id)}
+          onDownClick={() => handleDownClick(id)}
+          onUpClick={() => handleUpClick(id)}
+          onRemoveClick={() => handleRemoveClick(id)}
+        />
+      );
+      return (
+        <Reorder.Item key={card.id} value={card}>
+          {value}
```
jiejasonliu, line 193: nit: Would it better to get rid of the variable and put the jsx here? I found myself looking up and down

also framer motion is so cool 😮

src/components/drop-down/drop-down-choice.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+interface DropDownChoiceProps {
+  variant?: 'dark' | 'light';
+  className?: string;
+  bubbleOnClickEvent?: boolean;
+  onClick: (event: React.MouseEvent) => void;
+}
+
+export const DropDownChoice = ({
```
jiejasonliu, line 9: Is this used anywhere?

jiejasonliu (commented):
Few bookkeeping things, just want your feedback before approving. 

Regarding svgs as icon components, I think the way you have it is great -- and the correct choice for scalability. It wouldn't be feasible to have multiple svgs just for a simple color tweak.

src/app.tsx
derricklee43: This was just a makeshift parent component to render the flashcards as a group. I am working on the real parent right now and will have it fixed

src/app.tsx
derricklee43: Yeah, it's the same kind of makeshift parent component again, haha. I'll have it fixed with the new one

src/components/drop-down/drop-down-choice.tsx
derricklee43: I forgot to remove it, haha. I'll fix that right now

src/app.tsx
derricklee43: huh, good to know, haha.

src/components/flashcard/card-face/card-face.tsx
derricklee43: I branched off this branch. I will update it there and put it in the next PR.

PR #14, opened 2022-03-30
added header and sidebar

derricklee43:
- I am not sure how we want to handle navigation yet 
  -  For the demo, I assumed we can just have a state bool `showDecks` and this determines if we show the decks or cards of a deck page.
  -  clicking `flashcard decks` in the sidebar will set it to true, and we could set it up so clicking edit on a deck could set it to false
  - We can rework the links later to be a lot more sound
- There is also no handlers set up in the search bar in the header
  - We can add the handlers and mock data for the demo

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -8,6 +8,7 @@
   font-size: 15px;
   box-sizing: border-box;
   border: 1px solid transparent;
+  white-space: nowrap;
 
   &.dark {
     border-color: $light-blue;
--- a/src/components/sidebar/sidebar.tsx
+++ b/src/components/sidebar/sidebar.tsx
@@ -1,0 +1,23 @@
+import './sidebar.scss';
+import React from 'react';
+interface SidebarProps {
+  onFlashcardDecksClick: (event: React.MouseEvent) => void;
+}
+export const Sidebar = ({ onFlashcardDecksClick }: SidebarProps) => {
+  return (
+    <div className="c-sidebar">
+      <label className="c-title">
+        echo<span>Study</span>
+      </label>
+      <hr className="c-sidebar-divider" />
+      <ul>
+        <li>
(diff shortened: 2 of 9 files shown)


src/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,23 @@
+import './sidebar.scss';
+import React from 'react';
+interface SidebarProps {
+  onFlashcardDecksClick: (event: React.MouseEvent) => void;
+}
+export const Sidebar = ({ onFlashcardDecksClick }: SidebarProps) => {
+  return (
+    <div className="c-sidebar">
+      <label className="c-title">
```
jiejasonliu, line 9: Let's make this `c-sidebar-title` to avoid future potential conflicts.

src/components/button/button.scss
```diff
@@ -8,6 +8,7 @@
   font-size: 15px;
   box-sizing: border-box;
   border: 1px solid transparent;
+  white-space: nowrap;
```
jiejasonliu, line 11: Would the ellipsis mixin be something we want here instead?

jiejasonliu:
LGTM, definitely not the route in the long term but for demo purposes this is great.

jiejasonliu (changes requested):

src/components/button/button.scss
derricklee43: This just prevents the button from breaking into two lines, so I don't have to specify a width. But if we want to have ellipses for buttons, I can go back and add a width.

src/components/sidebar/sidebar.tsx
derricklee43: good catch. fixed

src/components/button/button.scss
jiejasonliu: I think we're okay without for now

jiejasonliu (approved):

PR #15, opened 2022-03-31
Add deck editor

derricklee43:
It is still kind of a mess🥲
I have logged most of the issues in trello, and left todo's, but I plan on going over it a lot more thoroughly after the demo and adding testing. 

https://user-images.githubusercontent.com/39753553/160962009-c360d478-8c94-4514-a580-fca2a2426fd2.mov


src/app.scss
```diff
@@ -6,14 +6,19 @@
 }
 
 .App {
+  .text-areas {
```
jiejasonliu, line 9: This is duplicated, probably some auto-merge trickery here -- probably safe to leave this here if we want to merge asap tho

src/components/deck-editor/meta-data-editor/meta-data-editor.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { Deck } from '../../../models/deck';
+import { Button } from '../../button/button';
+import { TextBox } from '../../text-box/text-box';
+import { BubbleDivider } from '../../bubble-divider/bubble-divider';
+import { DropDown, DropDownOption } from '../../drop-down/drop-down';
+import { Language } from '../../../models/card-content';
+import { TextArea } from '../../text-area/text-area';
+
+// Todo: maybe change Language to be enum instead of type
+enum Languages {
```
jiejasonliu, line 12: For future reference, we have this as a type in `card-content.ts`

src/components/drop-down/drop-down-choice.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+interface DropDownChoiceProps {
+  variant?: 'dark' | 'light';
+  className?: string;
+  bubbleOnClickEvent?: boolean;
+  onClick: (event: React.MouseEvent) => void;
+}
+
+export const DropDownChoice = ({
```
jiejasonliu, line 9: We don't need this right?

jiejasonliu (approved):
LGTM, if it's not too much work you can address some comments, otherwise you're free to merge since we're gonna be fixing a lot of this up anyway

src/components/deck-editor/meta-data-editor/meta-data-editor.tsx
derricklee43: Yeah, that's what I meant, haha.

But if we need to keep it is a type, we could create a Record<Language, string> object in the card-content.ts file and update them together

src/app.scss
derricklee43: haha, whoops. yep, I got it removed

src/components/deck-editor/meta-data-editor/meta-data-editor.tsx
jiejasonliu: I think at that point an enum is probably better

PR #17, opened 2022-04-06
fixed cursor stylings

derricklee43:
- updated the hover cursor for buttons, dropdown, search bars, and flashcards. 
- wired <Header /> to accept decks to insert into the searchbar


jiejasonliu (approved):
love me some polish! LG™

jiejasonliu:
kinda unrelated, but I notice that on smaller buttons, it's kinda hard to see the hover color change -- not really a problem but something to keep an eye out

derricklee43:
Yeah, we could probably add some padding to make it look better.

PR #19, opened 2022-04-08
Text area dark variant and animate label as legend

jiejasonliu:
- dark variant added to text-area component
- animate label as legend via `animateLabelAsLegend` prop (default: `true`)
- cleaned up some unused imports

text-area-legend-and-dark-variant


derricklee43 (approved):
LGTM!

PR #20, opened 2022-04-08
Inject esbuild-loader using craco to speedup compile times

jiejasonliu:
Forgot I had this in a stale branch from a month ago, test if this works on your machine and we can merge it

- use `esbuild-loader` instead of `babel-loader` 
- `esbuild` is a JS bundler written in Golang (fast bc it isn't interpreted)
- on my PC, it speeds up builds (`npm start` or `npm run build`) and hot-reload by 50%

image


derricklee43:
LGTM!

PR #25, opened 2022-04-09
Fix bubble divider overflow breaking children layout

jiejasonliu:
just a fix for a little bug i found while messing around, it made the dropdown cut off

- `overflow: hidden` causes absolute children to not break out
- solution: animate `overflow: hidden` and restore it on `transitionEnd` of visible
  - why `transitionEnd`? restoring it at the beginning will make it show before the animation expands to the area

**before:**
animate-overflow-before

**after:**
animate-overflow-after


derricklee43 (approved):
LGTM!

PR #26, opened 2022-04-10
Add page navigation using react-router-dom

jiejasonliu:
yayyy, we actually have page navigation!!! 🎉 

- entire app route definitions are in `routes.tsx`
- sidebar links defined in `components/sidebar/sidebar-routes.tsx`
- removed a lot of callback props, routing handled in page components
  - (e.g.) deck editor page now pulls `deckId` from the query param (/deck-editor/:deckId)
- prevent page navigation in deck-editor using workaround hook: `usePrompt(message: string, when: boolean)`

**side note...** what do we _actually_ want `go back` to do in the deck editor? go to home page, or actually go to the prev page? because this can cause issues if the user gets to the page directly instead of through our app.

https://user-images.githubusercontent.com/29434693/162596695-39a1c497-65db-4226-9f17-e3432b211d14.mp4

### DIFF
--- a/src/components/sidebar/sidebar-routes.tsx
+++ b/src/components/sidebar/sidebar-routes.tsx
@@ -1,0 +1,23 @@
+interface SidebarRouteItemProps {
+  name: string;
+  route: string;
+}
+
+export const SIDEBAR_ROUTE_ITEMS: SidebarRouteItemProps[] = [
+  {
+    name: 'flashcard decks',
+    route: '/', // todo: actually probably have a homepage
+  },
+  {
+    name: 'create',
+    route: '/deck-editor/0', // todo: maybe different route without any params (:deckId)
+  },
+  {
+    name: 'study',
+    route: '/study',
+  },
+  {
+    name: 'search',
+    route: '/search',
+  },
+];
--- a/src/components/sidebar/sidebar-routes.tsx
+++ b/src/components/sidebar/sidebar-routes.tsx
@@ -3,10 +3,12 @@ interface SidebarRouteItemProps {
   route: string;
 }
 
+// todo: actually probably have a homepage and maybe make `echoStudy` go back to home
+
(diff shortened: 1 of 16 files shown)


src/components/sidebar/sidebar-routes.tsx
```diff
@@ -0,0 +1,23 @@
+interface SidebarRouteItemProps {
+  name: string;
+  route: string;
+}
+
+export const SIDEBAR_ROUTE_ITEMS: SidebarRouteItemProps[] = [
+  {
+    name: 'flashcard decks',
+    route: '/', // todo: actually probably have a homepage
```
derricklee43, line 9: haha, yeah a real homepage would be nice😆

src/pages/deck-editor/deck-editor.tsx
```diff
@@ -110,6 +117,12 @@ export const DeckEditorPage = ({
   }
 
   function handleDeleteDeckClick(event: React.MouseEvent) {
-    onGoBackClick(event);
+    onGoBackClickHandler();
+  }
+
+  function onGoBackClickHandler() {
+    // todo: discuss what 'go back' should actually do
+    // can use navigate(-1) to go back a page, but this can have issues
```
derricklee43, line 125: if we have trouble popping the history, we could always change the text to be a fixed page

derricklee43 (approved):
LGTM!

src/pages/deck-editor/deck-editor.tsx
jiejasonliu: not really an issue, peep the side note, just potentially confusing

per your fixed page suggestion, we _could_ just send them back to the deck page, that's not a bad idea -- but maybe i'm thinking too much and just popping the history stack is good enough. thoughts?

derricklee43: Yeah, I saw we have it fixed for now. We could have something like `back to decks`. Then we can always change later when we have more inward routes from other pages

PR #27, opened 2022-04-10
Add drop down testing

derricklee43:
- make options tab-able 
- use Fade animation for the dropdown menu
- update Fade to toggle fade in
- add testing to dropdown

https://user-images.githubusercontent.com/39753553/162597735-cc0cc9f8-44b8-48a3-8085-ee6a1dcb883a.mov

[EDIT]
Summary of Additional Changes:
- Broke DropDownOptions into a separate component
- DropDown and SearchBar are now consumers of DropDownOptions
- Fixed SearchBar Testing with the new changes
- Fixed styling of DropDownOptions
  - When the component has a single option, only the bottom of the element had rounded corners 
- Updated the style of SearchBar to have rounded corners
  - In the mocks it was square, but I am now thinking rounded corners would be more consistent. What do you think?
- Added focus trap to SearchBar when options are shown
- Made the cancel button in the SearchBar focusable


jiejasonliu (approved):
looks great! can you do this for the dropdown in `search-bar.tsx` as well?

i'm kinda feeling i put us in an anti-pattern here since we're basically using the same logic/styles in two components.

not really the scope of your PR unless you want to, but maybe we should push the drop-down contents into a full blown component

derricklee43:
Yeah, I think breaking it into a separate component is a good idea. I'll get that done!

jiejasonliu (approved):
> In the mocks it was square, but I am now thinking rounded corners would be more consistent. What do you think?

total fan of this, makes things seem a little more consistent.
thanks for doing all of this, really good improvements here!

PR #28, opened 2022-04-10
created a stacking context for the page content

derricklee43:
- by creating a stacking context, we don't need to worry about elements occluding the header no matter their z-index


jiejasonliu (approved):
great solution, LGTM!

PR #29, opened 2022-04-10
derives lanaguge type from string list

derricklee43:
- we are now deriving the `Language` type from a read-only array
  -  `Language` is renamed to `DeckLanguage`
  - allows us to iterate over the array for the dropdown options
-  created new type: `CardLanguage`
  - `CardLanguage` is derived from `DeckLanguage` with the added `Default` option
  - `Default` allows to not have to cascade changes (and decide which will not be overridden)


src/models/card-content.ts
```diff
@@ -1,7 +1,11 @@
-export type Language = 'English' | 'Spanish' | 'German' | 'Japanese';
+import { DeckLanguages } from './deck';
+
+const cardLanguages = ['Default', ...DeckLanguages] as const;
```
jiejasonliu, line 3: can we export this and make it `CardLanguages`?

jiejasonliu (approved):
good changes here and just to make sure I understand: 

if a card front has `Default` and the deck front is `English`, then the card implicitly is `English`?

derricklee43:
Yeah, that's what I was thinking. I am ok with getting rid of `Default` I just wasn't sure how to handle the default language of the deck changing. Like would we reset all card languages, or the just the ones that are set manually? 

ikd, what do you think?

PR #30, opened 2022-04-12
Add api clients (part 1)

jiejasonliu:
let me know if this is what you had in mind for these api clients

i'd also like some feedback on if you like the formatting i went with! 
i wasn't too sure on how to best organize things so we know what's available in the client

- idea is to have the interface and we implement the methods as we need to use them
  - cards client 
  - decks client
  - fetch wrapper

* usage examples:
```ts
// with .then (if you're a boomer)
const decksClient = useDecksClient();
decksClient.addDeck(...).then((newDeck) => ...);

// with async and await
const decksClient = useDecksClient();
const newDeck = await decksClient.addDeck(...);

// with destructuring assignment
const { addDeck } = useDecksClient();
const newDeck = await addDeck(...);
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,8 +2,6 @@
     "editor.codeActionsOnSave": {
         "source.fixAll.eslint": true
     },
-    "[scss]": {
-        "editor.tabSize": 2,
-        "editor.formatOnSave": true
-    },
+    "editor.tabSize": 2,
+    "editor.formatOnSave": true
 }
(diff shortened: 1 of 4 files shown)


derricklee43:
I really like the formatting! I think it was a good idea to break apart the decks and the cards. Then we would also add another file for the logins and logouts, right?

LGTM!

derricklee43:
I really like the formatting! I think it was a good idea to break apart the decks and the cards. Then we would also add another file for the logins and logouts, right?

LGTM!

derricklee43 (approved):

jiejasonliu:
ok awesome! yeah exactly, something like `use-user-client.ts` with those functions you mentioned

PR #31, opened 2022-04-14
Basic api requests and fetch decks for flashcard-decks page

jiejasonliu:
- Implement fetch wrapper with basic functionality, no JWT/auth tokens yet
- Changes to `useDeckClient` --> implement `getAllDecks`, `getPublicDecks`
---
- Cannot actually hit the endpoints in dev environment due to CORS
  - To test, I copied the response from the endpoint
  - We probably want to fix CORS on the backend and test if it actually works before merging
---
fetched-decks-from-api

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+ECHOSTUDY_HOME=https://echostudy.com
+ECHOSTUDY_API_URL=https://api.echostudy.com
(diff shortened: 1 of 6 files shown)


derricklee43 (approved):
everything looks good to me!

PR #32, opened 2022-04-14
added useDeckEditor hook

derricklee43:
summary of changes: 

- combined all state changes in `DeckEditor` into a `useReducer`
- created a custom hook to wrap calls and state of new `useReducer`
  - now business logic and bookkeeping are in an isolated module
-  added testing to `DeckEditor`
  - additional testing is needed
    - fetching deck
    - deleting deck
    - saving deck
  - modified `Deck` to have `DeckMetaData` as a member

### DIFF
--- a/src/components/flashcard-set/flashcard-set.tsx
+++ b/src/components/flashcard-set/flashcard-set.tsx
@@ -9,19 +9,23 @@ interface FlashcardSetProps {
   variant?: 'readonly' | 'editable';
   className?: string;
   cards: Card[];
-  onCardsChange?: (cards: Card[]) => void;
+  onDeleteCardClick?: (card: Card) => void;
+  onCardReorder?: (cards: Card[]) => void;
+  onCardChange?: (card: Card) => void;
 }
 
 export const FlashcardSet = ({
   variant = 'editable',
   className = '',
   cards,
-  onCardsChange,
+  onCardReorder,
+  onDeleteCardClick,
+  onCardChange,
 }: FlashcardSetProps) => {
   const [activeCard, setActiveCard] = useState('');
   return (
     <div className={`c-flashcard-set ${className}`}>
-      <Reorder.Group axis="y" values={cards} onReorder={handleReorder}>
+      <Reorder.Group axis="y" values={cards} onReorder={handleCardReorder}>
         {getCards()}
       </Reorder.Group>
     </div>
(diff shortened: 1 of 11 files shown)


src/hooks/use-deck-editor.ts
```diff
@@ -0,0 +1,193 @@
...
+      if (metaData === undefined) throw new Error('action.metaData must not be undefined');
+
+      return { ...state, currentDeck: { ...currentDeck, metaData }, hasUnsavedChanges: true };
+
+    case DECK_REDUCER_TYPE.UPDATE_CARD:
+      if (card === undefined) throw new Error('action.card must not be undefined');
+
+      return {
+        ...state,
+        addedCards: card.id ? addCardToMap(card, addedCards) : addedCards,
```
jiejasonliu, line 145: i believe the logic is reversed here? 

if a card was updated and has an id, that means the card came from the backend; so we shouldn't add it to the CardsMap.

src/hooks/use-deck-editor.ts
```diff
@@ -0,0 +1,193 @@
...
+        updatedCards: !card.id ? addCardToMap(card, updatedCards) : updatedCards,
```
jiejasonliu, line 146: same idea here

src/components/flashcard-set/flashcard-set.tsx
```diff
@@ -30,16 +34,16 @@ export const FlashcardSet = ({
   function getCards() {
     return cards.map((card, index) => {
       return (
-        <Reorder.Item key={card.id ?? card.key} value={card}>
+        <Reorder.Item key={card.key} value={card}>
```
jiejasonliu, line 37: this is assuming that when we pull cards from the API, we set key = id?

in other words, the invariant:
- if new card on front end:
  - id: undefined 
  - key: uuidv4()
- if card fetched from API:
  - id = key

derricklee43: I was assuming that they would always remain separate. So every card would have a `key` that is a constant uuid that we create after fetching the cards.

When we add a card to the database with a fetch call, we would then only update the id on that card.

src/hooks/use-deck-editor.ts
derricklee43: Thanks for the callout. I made a last-minute change to check if new and get the logic flipped 😅. Fixed now.

src/components/flashcard-set/flashcard-set.tsx
jiejasonliu: ah I see, if we're okay with the DOM being completely reset when `Save` is clicked, let's proceed
especially not a concern if we're going the route with **not** refetching on save as you mentioned in Discord

jiejasonliu (approved):

src/components/flashcard-set/flashcard-set.tsx
derricklee43: Wait, it would be completely reset? I thought if we didn't re-fetch the cards, only the newly added cards would be re-rendered, which would be the same either way, right?

jiejasonliu: oh yeah, that's what i meant. not a concern if we're not refetching anymore

PR #35, opened 2022-04-15
Decks recoil state

jiejasonliu:
started the basis of using `Recoil` with this PR

**summary of Recoil and how it's used here:**
- `atoms` are the simplest form of state (key, value) pair
   - access using `useRecoilState` or if you only need the setter/getter `useSetRecoilState`/`useRecoilValue` respectively
- `selectors` are derived states, they update when atoms they depend on change
   - for example, sorted decks should update when the user deck **_or_** sort option selected changes
   - (i.e.) in this PR,  `userDecksSortedState` depends on the atoms: `userDecksState` and `userDecksSortedState`
- just like React's `useState`, whenever one of the recoil state changes, the component that consumes it is rerendered

**summary of changes:**
- add extra properties to `DeckMetadata`, we need this to sort by last created, say... maybe we want a _last updated_ option too?
- user decks are a recoil state, so is the selected sorting option; however, to view the sorted decks, we should subscribe to the sorted user decks and NOT user decks
  - this is extremely useful if we want to persist the sorting state in a future component aka the search page
- search bar data consumes the user decks recoil state, so the dropdown should update accordingly

https://user-images.githubusercontent.com/29434693/163634296-d506bc62-eb26-4e9c-963c-f1d020cd3d3c.mp4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "react-dom": "^17.0.2",
         "react-router-dom": "^6.3.0",
         "react-scripts": "5.0.0",
+        "recoil": "^0.7.2",
         "sass": "^1.49.9",
         "typescript": "^4.5.5",
         "uuid": "^8.3.2",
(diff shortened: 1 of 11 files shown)


derricklee43 (approved):
everything looks good to me! I think adding that last updated sort option is a good idea. We could also add last studied; I don't think there really is a limit to how many options we want to provide.

jiejasonliu:
cool! i'll add those options and if we need more in the future, they'll be easy to add

PR #36, opened 2022-04-16
added deck and cards fetching

derricklee43:
Summary of changes:

- Added deck and cards fetching
- Broke page navigation and fetch calls into parent component: `EditDeckPage`
- created a `setDeck` method on `useDeckEditor` aloowing us reset the component even when not unmounted
- Update routes
  - `\edit-deck\:deck-id` a route to an existing deck
  - `\create-deck` a route to a new deck
  - `EditDeckPage` handles both routes and fetches deck if route is `edit-deck`
  - moved all paths to an enum for an easier reference between pages

I removed all navigation and fetching out of the `DeckEditor` to keep the complexity of that component down

https://user-images.githubusercontent.com/39753553/163681192-604bd689-7110-4994-973d-ec79582d3232.mov


src/hooks/api/use-cards-client.ts
```diff
@@ -46,8 +51,26 @@ export function useCardsClient() {
...
+        text: obj['ftext'],
+        audio: obj['faud'],
+      };
+      card.back = {
+        language: obj['blang'],
+        text: obj['btext'],
+        audio: obj['baud'],
+      };
+      return card;
+    });
```
derricklee43, line 71: You are definitely right about needing a method to convert to JSON. If we convert the Deck to a class and add a static method, we wouldn't have to change the syntax of creating and copying decks, I believe.

src/pages/edit-deck-page/edit-deck-page.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }, [location]);
+
+  if (deck === undefined) {
+    return <LoadingPage label="loading deck..." />;
+  }
+
+  return (
+    <Fade className="edit-deck-page">
+      <DeckEditor
+        isNewDeck={false}
```
jiejasonliu, line 38: we'd want to pass `isNewDeck` here right?

src/pages/edit-deck-page/edit-deck-page.tsx
```diff
@@ -0,0 +1,70 @@
...
+        onCreateDeckClick={handleCreateDeckClick}
+        onDeleteDeckClick={handleDeleteDeckClick}
+        onGoBackClick={navigateBackToDecks}
+      />
+    </Fade>
+  );
+
+  async function fetchDeckAndRefresh() {
+    if (deckId === undefined) {
+      throw new Error('deckId cannot be undefined'); // Todo: maybe route to a 404 page??
```
jiejasonliu, line 49: definitely something we should do, any thoughts on what our 404 should look like?

src/routes.tsx
```diff
@@ -1,16 +1,26 @@
 import React from 'react';
 import { Routes, Route } from 'react-router-dom';
-import { DeckEditorPage } from './pages/deck-editor/deck-editor';
+import { EditDeckPage } from './pages/edit-deck-page/edit-deck-page';
 import { FlashcardDecksPage } from './pages/decks/flashcard-decks';
 
+export const enum paths {
```
jiejasonliu, line 6: good improvement, i like this 👍

jiejasonliu (approved):

src/hooks/api/use-cards-client.ts
jiejasonliu: yeah that would definitely make things a lot cleaner

jiejasonliu:
i've also noticed it gets quite confusing when there's a lot of cards and `+ new card` is clicked
maybe we want to [smooth] scroll to the top of the page when it's clicked?

src/pages/edit-deck-page/edit-deck-page.tsx
derricklee43: yep, thanks for catching that😅

derricklee43:
Yeah, that's I good idea. I'll add some smooth scrolling in another PR

PR #38, opened 2022-04-16
Add view deck page

derricklee43:
NOTE: This branches off `add-deck-and-card-fetching` and is merged with `toggle-speaker-and-kebab`, so we will just merge this last

Summary of changes
- add ViewDeck page
- added routes to support view deck page
- updated default variant of flashcard set to be `readonly`
- updated DeckTile to have `view` instead of `edit` 

Screen Shot 2022-04-16 at 12 58 58 PM

Screen Shot 2022-04-16 at 1 00 19 PM

[EDITED DESIGN]
Screen Shot 2022-04-16 at 3 04 35 PM


jiejasonliu (approved):
overall approved but a question:

for our `BubbleDivider` component, does it support not rendering the button?
if yes, I think it'd look better as just the divider. if not, will we add functionality to do so, or would this be some other divider component (i.e. `LineDivider`), then again it would just be a single `hr` at that point...

derricklee43:
Yeah, it's not doing anything. You're right, a simple hr would be better here. I'll get that changed

derricklee43:
I think we will have a bit of redesign to do this page later on

PR #40, opened 2022-04-16
added smooth scrolling to bottom for new cards

derricklee43:
oof, that was a lot tougher than I was expecting lol

- added cards are now added to the bottom of the set
- automatically scroll to the added card if not in view
- automatically set a newly added card to be active

https://user-images.githubusercontent.com/39753553/163693703-066e4e70-80fd-4e0a-9388-e4bfe7899558.mov

### DIFF
--- a/src/components/flashcard/flashcard.tsx
+++ b/src/components/flashcard/flashcard.tsx
@@ -6,6 +6,8 @@ import { FlashcardArrowIcon } from '../../assets/icons/flashcard-arrow-icon/flas
 import { Button } from '../button/button';
 import { Card } from '../../models/card';
 import { CardContent } from '../../models/card-content';
+import { useEffect } from 'react';
+import { useRef } from 'react';
 
 interface FlashcardProps {
   card: Card;
@@ -30,8 +32,23 @@ export const Flashcard = ({
   onRemoveClick,
   onSpeakerClick,
 }: FlashcardProps) => {
+  const cardRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (cardRef.current === null || variant !== 'active') {
+      return;
+    }
+
+    if (!isCardTopInViewPort(cardRef.current)) {
+      cardRef.current.scrollIntoView({
+        behavior: 'smooth',
+        block: 'center',
+        inline: 'center',
+      });
+    }
+  }, [variant]);
+
   return (
-    <div className={`flashcard ${variant}`}>
+    <div ref={cardRef} className={`flashcard ${variant}`}>
       {variant !== 'readonly' && getLeftButtonStrip(variant)}
       <CardFace
         variant={variant}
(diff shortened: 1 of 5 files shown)


src/components/flashcard/flashcard.tsx
```diff
@@ -86,4 +103,9 @@ export const Flashcard = ({
   function handleBackFaceChange(cardContent: CardContent) {
     onCardChange?.({ ...card, back: cardContent });
   }
+
+  function isCardTopInViewPort(cardDiv: HTMLDivElement) {
+    const { top } = cardDiv.getBoundingClientRect();
+    return top <= (window.innerHeight || document.documentElement.clientHeight);
```
jiejasonliu, line 109: correctness wise, shouldn't we also check if `bottom > 0`?

but i guess it technically doesn't matter since new cards appear at the bottom

jiejasonliu (approved):
thanks for figuring this out 😄

src/components/flashcard/flashcard.tsx
jiejasonliu: also logic wise, this would mean (and what you have currently) that this returns true if the card **overlaps** instead of **in**

correct logic would be `top >= 0` and `bottom <= (window.innerHeight || document.documentElement.clientHeight)`

derricklee43: I had it like this, but it was a little jarring because it scrolls to center, and often the cards are just a little off the screen, so it was too much movement. 

 I will try breaking the scroll into view into 2 cases: is too high and is too low. And it scrolls just enough to bring the entire card into view.

derricklee43: Jk, that did not work at all. It will scroll just enough to get into view, so it is occluded by the header on the top and occluded by the add card button on the bottom if we do that. 

The original logic is it will only scroll if a card becomes active when completely outside of the viewport. I assumed this should only happen on `new card add` but I can add the reverse logic for the top for completeness.

jiejasonliu: ah that makes perfect sense on that we only want to scroll if it's _completely_ out of the view (negation of any overlap)
left a comment on your most recent commit, once that's handled you can merge

sorry for the nits

src/components/flashcard/flashcard.tsx
```diff
@@ -105,7 +105,7 @@ export const Flashcard = ({
   }
 
   function isCardTopInViewPort(cardDiv: HTMLDivElement) {
-    const { top } = cardDiv.getBoundingClientRect();
-    return top <= (window.innerHeight || document.documentElement.clientHeight);
+    const { top, bottom } = cardDiv.getBoundingClientRect();
+    return bottom >= 0 || top <= (window.innerHeight || document.documentElement.clientHeight);
```
jiejasonliu, line 109: scroll wouldn't work anymore, this should be `&&` (since `bottom >= 0` is always true)

and while you're at it, can we rename this method to `isCardInViewport` to be more concise?

derricklee43: haha, it's been a long day, thanks for catching that😂

PR #41, opened 2022-04-17
Implement add delete in use decks client

derricklee43:
- implemented add and delete deck with the associated routing
- delete endpoint is a little scuffed, but we can figure it out after the demo

### DIFF
--- a/src/pages/edit-deck-page/edit-deck-page.tsx
+++ b/src/pages/edit-deck-page/edit-deck-page.tsx
@@ -15,7 +15,7 @@ export const EditDeckPage = () => {
   const { deckId } = useParams(); // via the param :deckId
   const location = useLocation();
   const navigate = useNavigate();
-  const { getDeckById } = useDecksClient();
+  const { getDeckById, deleteDeckById, addDeck } = useDecksClient();
   const { getCardsByDeckId } = useCardsClient();
 
   const isNewDeck = location.pathname === paths.createDeck;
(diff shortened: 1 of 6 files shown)


src/pages/edit-deck-page/edit-deck-page.tsx
```diff
@@ -53,14 +53,20 @@ export const EditDeckPage = () => {
...
+  async function handleCreateDeckClick(deck: Deck) {
+    const addedDeck = await addDeck(deck);
+    navigate(`${paths.deck}/${addedDeck.metaData.id}`);
   }
 
-  function handleDeleteDeckClick() {
-    // Todo: delete deck
+  async function handleDeleteDeckClick() {
+    if (!isNewDeck && deck !== undefined) {
+      // Todo: always throws. It looks like the response is bad but the deck is still deleted
```
jiejasonliu, line 63: oh! i'm 99.99% certain this is on me, delete requests don't have a response so this line would definitely throw:

`const data = response.json()` in `use-fetch-wrapper.ts:39`

you can fix this by changing it to:

```ts
const text = await response.text();
const data = text && JSON.parse(text);
```

jiejasonliu (approved):
i'm okay with merging this despite API changes coming in the future

we'll just need to be careful for that

src/pages/edit-deck-page/edit-deck-page.tsx
derricklee43: sweet, that fixed it

derricklee43:
Yeah, we can just leave it open until the API is fixed and we have it updated

derricklee43:
Ok, we should be good to go for these updated endpoints

jiejasonliu (approved):
LGTM, do you mind updating the comments w the correct endpoint info?

derricklee43:
yep, there are now updated

PR #42, opened 2022-04-17
Flesh out save method in use deck editor

derricklee43:
- makes API calls in custom hook
- once calls are complete, update the reducer
- removed repeated code for `save` and `update`

NEW CHANGES
- Button transition to saving state in deck editor
- Added isSaving state to useDeckEditor hook
  - A dispatch requests to reducer are ignored when isSave is true

### DIFF
--- a/src/animations/up-toggle.tsx
+++ b/src/animations/up-toggle.tsx
@@ -24,14 +24,11 @@ export const UpToggle = ({
     setIsFirstRender(false);
   }, []);
 
-  const yOffset = 30;
-  const activeVariant = { y: 0, opacity: 1 };
-  const inactiveVariant = isFirstRender ? activeVariant : { y: yOffset, opacity: 0 };
-
+  const inactiveVariant = { y: 30, opacity: 0 };
   const variants = {
-    active: activeVariant,
+    active: { y: 0, opacity: 1 },
     inactiveDefault: inactiveVariant,
-    inactiveAlternate: { ...inactiveVariant, y: yOffset * -1 },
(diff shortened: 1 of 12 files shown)


jiejasonliu:
how difficult would it be to introduce a loading state since save times can be unpredictable?

it'd be either through recoil states or introducing a saving state in the hook i think.

either way, doesn't have to be done in this PR, could be a post-demo thing. 

let's document it in trello if we're tackling this later though.

jiejasonliu (approved):

src/animations/up-toggle.tsx
```diff
@@ -0,0 +1,54 @@
...
+  className = '',
+  showDefault = true,
+  defaultContent,
+  alternateContent,
+}: UpToggleProps) => {
+  // prevents transition on mount
+  const [inactiveY, setInactiveY] = useState(0);
+  const [inactiveOpacity, setInactiveOpacity] = useState(1);
+
+  // enables transition after mount
```
jiejasonliu, line 23: for future reference, this can get messy if we have a lot of initial states we want to set

if it becomes too complex, i think a single `isFirstRender` state can simplify this since if `initial={false}` it goes to takes whatever is in `animate` to be default

jiejasonliu (approved):
LGTM really clean animation

src/animations/up-toggle.tsx
derricklee43: That is a way cleaner solution. I changed it to that. Thanks for the callout here.

PR #43, opened 2022-04-17
Add way too many tests

jiejasonliu:
add tests for every component, a couple components aren't finished... (i.e. header) 
but we should test that it displays properly at the minimum

---

* notable changes aside from tests
  - improve class names for `FlipTile`, `PageHeader`, `ProgressBar`
  - add `labelDelay` as a prop for `LoadingPage`
  - add `renderWithHistoryRouter` to easily test components that can change the URL 
    - as an example, see: `sidebar.test.tsx`
  - `handleUpClick` for `FlashcardSet` now wraps around properly when clicking on the first card (it use to just swap with last)
  
---

* also added `npm run coverage` to get a feeling of how well-tested we are
  - this may be misleading since this coverage includes _helpers_, _hooks_, and _models_

before:
test-coverage-before

after:
test-coverage-after

### DIFF
--- a/src/components/drop-down-options/drop-down-options.test.tsx
+++ b/src/components/drop-down-options/drop-down-options.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { DropDownOption, DropDownOptions } from './drop-down-options';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { noop } from '../../helpers/func';
 import {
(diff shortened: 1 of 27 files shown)


src/components/drop-down-options/drop-down-options.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+    expect(screen.queryByText(TEST_OPTIONS_SINGLE_VALUES[0])).toBeVisible();
+  });
+
+  it('should hide when component rerenders with show as false', () => {
+    const { rerender, unmount } = render(
+      <DropDownOptions show={true} options={TEST_OPTIONS_SINGLE} onOptionSelect={noop} />
+    );
+    expect(screen.queryByText(TEST_OPTIONS_SINGLE_VALUES[0])).toBeVisible();
+
+    unmount(); // i think this is needed since AnimatePresence keeps old component alive despite wait(...)
```
derricklee43, line 25: Have you tried using something like this:

`await waitForElementToBeRemoved(() => screen.queryByText(TEST_OPTIONS_SINGLE_VALUES[0]));`

I know this usually works for child components of `<AnimatePresence>` but I'm not sure if this is something else.

src/components/flashcard-set/flashcard-set.tsx
```diff
@@ -66,8 +66,13 @@ export const FlashcardSet = ({
 
   function handleUpClick(index: number) {
     const newCards = [...cards];
```
derricklee43, line 68: I really like this change. I was thinking the behavior was weird but didn't quite realize why😅

derricklee43 (approved):
Damn! That is a lot of tests!😆 Thanks for doing that; it all looks good to me!

src/components/drop-down-options/drop-down-options.test.tsx
jiejasonliu: that was exactly it, thanks for the suggestion!

src/components/flashcard-set/flashcard-set.tsx
jiejasonliu: perks of writing tests I guess, I was bashing my head for 30 minutes trying to find out why my test didn't work.
lo and behold, it turns out it just didn't do what we expected 😆

PR #44, opened 2022-04-20
added deck cover transition

derricklee43:
- added deck cover animation when reordered 
- added text color transition for `add new deck` card

https://user-images.githubusercontent.com/39753553/164149673-f2e8e315-db41-4c20-aeb0-793178ca62cd.mov


jiejasonliu (approved):
damnn framer-motion is so extensible, LGTM!

PR #46, opened 2022-04-20
added 404 page and update routes

derricklee43:
Screen Shot 2022-04-20 at 12 05 39 AM

### DIFF
--- a/src/pages/edit-deck-page/edit-deck-page.tsx
+++ b/src/pages/edit-deck-page/edit-deck-page.tsx
@@ -46,7 +46,7 @@ export const EditDeckPage = () => {
 
   async function fetchDeckAndRefresh() {
     if (deckId === undefined) {
-      throw new Error('deckId cannot be undefined'); // Todo: maybe route to a 404 page??
+      throw new Error('deckId cannot be undefined');
     }
     const deck = await getDeckById(deckId);
     deck.cards = await getCardsByDeckId(deckId);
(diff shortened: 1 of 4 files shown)


src/pages/edit-deck-page/edit-deck-page.tsx
```diff
@@ -46,7 +46,7 @@ export const EditDeckPage = () => {
 
   async function fetchDeckAndRefresh() {
     if (deckId === undefined) {
-      throw new Error('deckId cannot be undefined'); // Todo: maybe route to a 404 page??
```
jiejasonliu, line 49: did we want to start routing to this new page?

src/pages/not-found-page/not-found-page.tsx
```diff
@@ -0,0 +1,6 @@
+import './not-found-page.scss';
+import React from 'react';
+
+export const NotFoundPage = () => {
+  return <div className="not-found-page">oopsie daisy! Page not found...</div>;
```
jiejasonliu, line 5: 🤭💮! 📄🚫🔍...

jiejasonliu (approved):
l g t m 💯

src/pages/edit-deck-page/edit-deck-page.tsx
derricklee43: It automatically routes it to the page. I just kept it because the type checker was complaining. But I guess we could explicitly send the user to /404 or something.

jiejasonliu: ah I see! bc `/edit-deck` on its own is an invalid route

we'll probs do some better error checking in the future since our endpoints can fail
i think i'll help/work on improving that post demo though

PR #47, opened 2022-04-20
prevent save when title or desc 

derricklee43:
- prevent save when title or desc is empty
- hide speaker and placeholder when a card is empty"
- add empty card set placeholder message in view decks page
- added tests
Screen Shot 2022-04-20 at 12 31 33 AM
Screen Shot 2022-04-20 at 12 31 44 AM


jiejasonliu (approved):
lookin good!!

image

derricklee43:
awe hells yeah!

PR #51, opened 2022-07-30
added importing cards from files

derricklee43:
Summary of changes: 
 - Added new doc icon
 - Added testing for ImportCardsPopup
 - Added FileInput
   - Asynchronously reads .txt files and exports them as strings
   - Added testing for FileInput 
   - If we run out of things to do, we can add functionality to read .docx files
- Redesigned ImportCards Popup
  - Showing both the textArea and FileInput created an ambiguous user experience
  - (I am not a fan of having both `upload file` and `browse your files` buttons as they are redundant, but adding functionality to open the native file browser directly from the RadioButtonGroup component will take a bit of refactoring) 

https://user-images.githubusercontent.com/39753553/181933675-a7c79061-32dc-4570-8407-e21e423fe538.mov


src/components/file-input/file-reader.tsx
```diff
@@ -0,0 +1,18 @@
+export function readFile(file: File) {
+  return readTxtFile(file);
+}
+
+function readTxtFile(txtFile: File) {
+  return new Promise<string>((resolve, reject) => {
+    const fileReader = new FileReader();
+    fileReader.addEventListener('load', () => {
```
jiejasonliu, line 8: consider adding an event listener for `error` (or using .onerror) to reject the promise; I'm fairly certain `readAsText` doesn't throw any synchronous exceptions

src/pages/import-cards-popup/import-cards-popup.test.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react';
+import { noop } from '../../helpers/func';
+import userEvent from '@testing-library/user-event';
+import { ImportCardsPopup } from './import-cards-popup';
+import { Card } from '../../models/card';
+
+describe('ImportCardsPopup', () => {
+  let mockOnClose: jest.Mock<void, [Card[]]>;
```
jiejasonliu, line 9: are we going to add more tests using this?
if not, it makes more sense to live in the test that's using it

src/pages/import-cards-popup/import-cards-popup.tsx
```diff
@@ -50,26 +60,13 @@ export const ImportCardsPopup = ({ showPopup, onClose }: ImportCardsPopupProps)
...
-            <Button variant="light" onClick={noop} className="import-cards-popup-button">
-              browse your files
-            </Button>
-          </div>
-        </div>
+        <div className="import-cards-list-item-label">3. how to import your cards?</div>
+        <RadioButtonGroup
+          radioButtonOptions={importMethodOptions}
+          selectedButtonKey={importMethod}
+          onButtonSelect={setImportMethod}
```
jiejasonliu, line 67: do we also want to clear the `importedFile` if this changes?
(i.e. we change to the `copy and paste` and back to `upload file`)

src/pages/import-cards-popup/import-cards-popup.tsx
```diff
@@ -26,12 +25,23 @@ const cardSeparatorOptions: RadioButtonOptions = {
...
+
 export const ImportCardsPopup = ({ showPopup, onClose }: ImportCardsPopupProps) => {
   const [termDefSeparator, setTermDefSeparator] = useState('');
   const [cardSeparator, setCardSeparator] = useState('');
   const [textAreaValue, setTextAreaValue] = useState('');
+  const [importMethod, setImportMethod] = useState('');
+  const [importedFile, setImportedFile] = useState<ImportedFile>();
 
-  const isImportable = termDefSeparator && cardSeparator && textAreaValue.length !== 0;
+  const isImportable = isImportConfigValid();
```
jiejasonliu, line 44: for the future, did we want to something more than just disable the `import` button? perhaps some error text could be useful here

src/components/drop-down-options/drop-down-options.test.tsx
```diff
@@ -24,7 +24,7 @@ describe('Button', () => {
 
     rerender(<DropDownOptions show={false} options={TEST_OPTIONS_SINGLE} onOptionSelect={noop} />);
     await waitForElementToBeRemoved(() => screen.queryByText(TEST_OPTIONS_SINGLE_VALUES[0]), {
-      timeout: 500, // wrt `Fade` having transition duration of 0.2 (200ms)
+      timeout: 1000, // wrt `Fade` having transition duration of 0.2 (200ms)
```
jiejasonliu, line 27: 😕 async code be like

jiejasonliu (commented):
ooo overall awesome changes, had to pull myself back to remember half the stuff we did 😂 

on the note of the redundant buttons, I don't think refactoring the group is the correct choice.
it's either we go back to the route of having two choices appearing at once, or stick with this and polish the `upload file` choice. maybe we should create something that matches the native?

none-selected

file-selected

src/components/file-input/file-input.tsx
```diff
@@ -0,0 +1,81 @@
...
+  label,
+  importedFile,
+  onImportedFileChange,
+}: FileInputProps) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const id = uuidv4();
+  return (
+    <div className={`c-file-input ${className}`}>
+      {importedFile ? getSelectedFile() : getSelectFileButton()}
+      <input key={id} type="file" id={id} name="myfile" onChange={importFile} />
```
jiejasonliu, line 34: there's also an `accept` attribute that's supported in some browsers if we want to allow users to only pick specific files

derricklee43:
> ooo overall awesome changes, had to pull myself back to remember half the stuff we did 😂
> 
> on the note of the redundant buttons, I don't think refactoring the group is the correct choice. it's either we go back to the route of having two choices appearing at once, or stick with this and polish the `upload file` choice. maybe we should create something that matches the native?
> 
> none-selected
> 
> file-selected

Ok, I updated the upload file choice to match the screenshots you shared
Screen Shot 2022-08-21 at 7 24 36 PM

src/components/file-input/file-reader.tsx
derricklee43: haha, whoops. Good catch. I got that updated

src/pages/import-cards-popup/import-cards-popup.test.tsx
derricklee43: I was thinking in case we were to add more tests in the future. But I moved it into the test; we can always move it back out again if we end up adding some more, haha

src/pages/import-cards-popup/import-cards-popup.tsx
derricklee43: Yeah, I wasn't quite sure either way on this one, haha. I ended up clearing both the textbox and selected file whenever it is toggled.

src/pages/import-cards-popup/import-cards-popup.tsx
derricklee43: Yeah, giving more info is a good idea to me. I'll create a ticket for it and mark it as low priority and we can revisit it if we have time

src/components/drop-down-options/drop-down-options.test.tsx
derricklee43: :sob:

src/components/file-input/file-input.tsx
derricklee43: good catch. I added it; it's just .txt files (for now, we could always add more if we get bored)

src/components/drop-down-options/drop-down-options.test.tsx
jiejasonliu: yeah I'm still not convinced this is a good pattern, it's highly dependent on the frame-rate (so however much cpu load the test container has) due to `requestAnimationFrame`

i think i'm gonna audit this test and rewrite it better in another PR

jiejasonliu (approved):

PR #52, opened 2022-07-30
added multiline flashcards

derricklee43:
https://user-images.githubusercontent.com/39753553/185822823-1ea328a0-d1f6-4968-8a09-b2959371ad01.mov

Summary of changes:
- swapped single-line inputs with multiline textboxes
- disabled dragging from any part of the card
  - Cannot overload both dragging and highlighting within a card
  - (We could devise a dragging system when it is inactive, but I think that would be more confusing)
- Added a drag icon
  - I am thinking about removing and disabling drag altogether
    - Doesn't work on mobile
    - Is a little hacky (doesn't work well with the current animation library)


src/components/flashcard-set/flashcard-set.tsx
```diff
@@ -26,70 +29,87 @@ export const FlashcardSet = ({
...
         </Reorder.Item>
       );
     });
   }
 
-  function getCardVariant(cardKey: string) {
-    if (variant === 'readonly') return 'readonly';
-    return activeCardKey === cardKey ? 'active' : 'inactive';
+  function onDragStart(e: any, info: any, key: string) {
+    if (e.target.classList.contains(DRAG_HANDLE)) {
```
jiejasonliu, line 84: for my education, if the drag handle lives in the child component's class
how does `e.target.classList` ever contain the handle?

or does `e.target` change depending on where you started dragging -- so this would only set the dragging key if you begin on the icon?

jiejasonliu (approved):
good refactors, I like the readonly flashcard set. I always thought that `readonly` as an attribute was a little too powerful. 

one concern though, did we want users to be able to playback a card in the readonly set? maybe I'm misunderstanding the code, but it doesn't appear to have that logic

PR #53, opened 2022-08-23
fix: use fake timers to resolve flaky dropdown options test

jiejasonliu:
title; it also speeds up tests bc we no longer have to wait for `rAF` so win-win situation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5596,13 +5596,19 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001312",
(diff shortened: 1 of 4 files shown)


derricklee43 (approved):
looks good to me! I am still surprised you found that jest bug, lol

jiejasonliu:
> looks good to me! I am still surprised you found that jest bug, lol

about that haha, browsers actually store the delay as a 32-bit int for perf reasons so this is actually expected behavior 🤷

PR #55, opened 2022-08-27
added stricter typing to dropdown menu

derricklee43:
Added stricter typing to the dropdowns. 

onOptionSelect can now infer the type from the given options. 

For arrays declared as const for the options, the type is exactly one of the provided options. 

If the options type is not strict (e.g. string[]) the option type will also be not strict (string following the example).


jiejasonliu (approved):
omg nice! this is one of the things that I meant to do but totally forgot about over the summer

derricklee43:
Haha, there is just so much to do

PR #59, opened 2022-09-01
feat: fetch with auth-jwt headers

jiejasonliu:
extends our fetch-wrapper to prep for handling auth JWTs + some misc stuff

### useFetchWrapper
- now able to retry requests (by default once)
- if a request fails on the server (e.g. 401 due to invalid JWT), we have a place to refresh the token (currently a TODO)

### useUserClient / useLocalStorage
- an attempt to log the user in will serialize the token to local storage
  - default `localStorage` only allows strings, `useLocalStorage` abstraction allows objects
- the serialized token lives in a recoil atom
  - if token in local storage when app first loads, it will be passed to the atom as a init step

### miscellaneous things added
- object schema validator (type-guards for objects)
- custom jasmine matcher `.toMatchPredicate(predicate)` to chain result of expectation to a predicate
- tests for the schema validator and fetch wrapper retries

### DIFF
--- a/src/helpers/validator.ts
+++ b/src/helpers/validator.ts
@@ -1,0 +1,26 @@
+/**
+ * Creates a user-type guard for objects provided a simple key-to-type schema.
+ * Optional (?) types should not be provided in the simple schema.
+ *
+ * @param schema - (e.g.) { 'message': 'string', 'id': 'number' }
+ */
+export function objectSchemaSimple<T extends object>(schema: Record<string, string>) {
+  return (subject: unknown): subject is T => {
+    if (!isObject(subject)) {
+      return false;
+    }
+
+    return Object.entries(schema).every(([key, expectedType]) => {
+      const value = subject[key];
+      return value && typeof value === expectedType;
+    });
+  };
+}
+
+export function isObject(subject: unknown): subject is Record<string, unknown> {
+  return subject !== null && typeof subject === 'object' && !Array.isArray(subject);
+}
+
+export function isString(subject: unknown): subject is string {
+  return typeof subject === 'string';
+}
--- a/src/hooks/api/use-fetch-wrapper.ts
+++ b/src/hooks/api/use-fetch-wrapper.ts
@@ -48,6 +48,7 @@ export function useFetchWrapper(prependApiUrl?: string) {
   ): Promise<any> {
     const maybeContentTypeHeader = body ? { 'Content-Type': 'application/json' } : undefined;
     const resolvedUrl = prependApiUrl ? prependApiUrl + url : url;
+
     const response = await fetch(resolvedUrl, {
       method: method,
       headers: {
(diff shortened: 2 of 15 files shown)


src/helpers/validator.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Creates a user-type guard for objects provided a simple key-to-type schema.
+ * Optional (?) types should not be provided in the simple schema.
+ *
+ * @param schema - (e.g.) { 'message': 'string', 'id': 'number' }
+ */
+export function objectSchemaSimple<T extends object>(schema: Record<string, string>) {
```
derricklee43, line 7: oh wow, this is dope. Took me a bit to figure out how it works, lol

src/helpers/validator.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Creates a user-type guard for objects provided a simple key-to-type schema.
+ * Optional (?) types should not be provided in the simple schema.
```
derricklee43, line 3: Are we planning on adding another method for type validation for other optional keys?

src/hooks/api/use-fetch-wrapper.ts
```diff
@@ -15,41 +30,78 @@ export function useFetchWrapper(prependApiUrl?: string) {
...
-  async function handleResponse(response: Response) {
-    const text = await response.text();
-    const data = text && JSON.parse(text);
+  async function _retryFetch(
+    url: string,
+    method: string,
+    body: object | undefined,
+    retries = 1
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ): Promise<any> {
```
derricklee43, line 48: Could we type this as unknown or will that break things?

src/hooks/use-local-storage.ts
```diff
@@ -0,0 +1,56 @@
...
+    const data = localStorage.getItem(key);
+    return data ? (JSON.parse(data) as T) : undefined;
+  }
+
+  function upsert(key: string, value: string | object): boolean {
+    try {
+      // value was an object
+      if (isObject(value)) {
+        const serializedValue = JSON.stringify(value);
+        localStorage.setItem(key, serializedValue);
```
derricklee43, line 33: Damn, I didn't realize we could write to storage this easily, haha

derricklee43 (approved):
This is really cool; I am glad you were able to figure this all out. LGTM!

src/hooks/api/use-fetch-wrapper.ts
jiejasonliu: we can and that would be the best thing forward, it just means we need to create object schemas for the response we get back on API calls

might be worth for a future story, because being honest, the API is changing a lot

src/helpers/validator.ts
jiejasonliu: i would like to, except that inherently an optional key of type T is `T | undefined`
a different type of validator could be instead of `key: 'type'` it would be `key: predicate`
this way we can support chaining (e.g. `key: optionalKey(isString)` would match `key?: string`)

but I'd if we want better validator, it might be worth just finding a battle-tested validator library instead of making one ourselves

derricklee43: haha, yeah. You're right, I'd imagine that this is a common problem

src/hooks/api/use-fetch-wrapper.ts
derricklee43: oh, I forgot about that. Yeah, nevermind then, haha

PR #61, opened 2022-09-02
added ResourceLoader

derricklee43:
## Summary of Changes
- Consolidated Page Loading Logic
- Consolidated Deck Fetching Logic
- Consolidated Refresh Logic
- Parameterized ResourceLoader to watch any one parameter in the URL


src/components/resource-loader/resource-loader.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const { getCardsByDeckId } = useCardsClient();
+
+  if (deckId === undefined) {
+    if (allowUndefinedDeckId) {
+      return createNewDeck();
+    }
+    throw Error('deckId cannot be undefined');
+  }
+
+  const deck = await getDeckById(deckId);
```
jiejasonliu, line 51: since these two promises are independent, we can double the perf with:
```ts
  const [deck, cards] = await Promise.all([getDeckById(deckId), getCardsByDeckId(deckId)]);
  deck.cards = cards;
```

jiejasonliu (approved):
good changes, with this change, we could unify a way to abort ongoing fetches/audio loads when the component is destroyed

i'm tracking that in trello and i'll take that on

src/components/resource-loader/resource-loader.tsx
derricklee43: ooof, that is nice. I forgot that was a thing, lol.

PR #63, opened 2022-09-03
fix: add base href to avoid url-rewrite using resource url as base

jiejasonliu:
an example rewrite of `/decks/1` -> `/` should now serve the right static files


derricklee43 (approved):
lgmt

PR #64, opened 2022-09-03
switched to promises instead of callbacks

derricklee43:


src/hooks/use-play-card-audio.ts
```diff
@@ -100,7 +77,16 @@ export function usePlayCardAudio() {
 
   function playAudio(audio: HTMLAudioElement) {
     setActiveAudio(audio);
-    chainToAudio(() => setActiveAudio(undefined), audio);
-    audio.play();
+    return new Promise<void>((resolve, reject) => {
+      audio.addEventListener('ended', () => resolve(), { once: true });
```
jiejasonliu, line 81: when the component is destroyed, this promise may or may not be GC'd depending on how the browser/React holds onto references, so it will likely keep playing.

a quick solution I came up with is this, but there's probably a better solution if you can come up w one
```ts
// keeps track of whether this component is still mounted
  const mountedRef = useRef<boolean>(false);
  useEffect(() => {
    mountedRef.current = true;
    return () => {
      mountedRef.current = false;
    };
  }, []);

...

audio.addEventListener('ended', () => mountedRef.current && resolve(), { once: true });
audio.addEventListener('error', () => mountedRef.current && reject('unable to play audio'), { once: true });
```

if going for this solution, it might even make sense to wrap that ref logic in a custom hook, so it can be one line
`const isComponentMounted = useComponentMountedState()` or something 🤷

jiejasonliu (changes requested):

src/hooks/use-play-card-audio.ts
derricklee43: Yeah, that's a good idea. I noticed that; it keeps playing after you navigate away. I added this

```
  useEffect(() => {
    return () => clearAudio();
  }, []);
```
where
```
  function clearAudio() {
    activeAudioRef.current?.pause();
    clearTimer();
  }
```

I've tested this and it pauses the audio immediately when a user leaves instead of just canceling the next queued audio. I think this would work, but I think you know this promise timing stuff better than I do. 

Is this good, or should we combine snippets by using `mountedRef` and call pause on the current audio?

jiejasonliu: this is good! this works because the timer is cleared which will release the reference, nice 👍

jiejasonliu (approved):

src/hooks/use-play-card-audio.ts
derricklee43: cool!

PR #65, opened 2022-09-03
fix: resolve invalid hook call in resource fetchers

jiejasonliu:
had a bit of a merge conflict between the JWT fetch wrapper and resource fetchers

the JWT PR added hooks to the fetch wrapper which caused a bug in the resource fetcher PR to surface.
specifically, the Rule of Hooks: [React doesn't allow hooks to be called outside of functional components / custom hooks](https://reactjs.org/docs/hooks-rules.html)

---

### Solution
- make the resource fetchers act like `custom hooks`! where logic can be lazily evaluated, but hooks consumed immediately
  - signature of resource fetchers are now `() => Promise<T>` 
- resource loader immediately invokes the resource fetcher for hooks, but the actual logic is wrapped in a closure to be invoked later

### DIFF
--- a/src/components/resource-loader/resource-loader.tsx
+++ b/src/components/resource-loader/resource-loader.tsx
@@ -6,19 +6,20 @@ import { createNewDeck } from '../../models/deck';
 import { LoadingPage } from '../loading-page/loading-page';
 
 interface ResourceLoaderProps<T> {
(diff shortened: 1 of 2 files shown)


derricklee43:
oh, haha whoops. These changes look good to me!

PR #66, opened 2022-09-04
feat: lazily load audio files

jiejasonliu:
# note to self: please rebase before merging to `main`
# note to derrick: i rebased off your branch so only review the last commit

### Loading all audio files when viewing/studying a deck is inefficient.
  - users will likely not listen to all audio files nor even listen to any audio files at all
instead, we lazily load audio files when needed using the new `LazyAudio` wrapper 

**(read header comment in `lazy-audio.ts`, mini architecture discussion)**

### Other improvements
- audio pauses and resumes correctly in a study session (need to reset active audio while waiting)
- in the deck viewer, starting another audio will stop any ongoing audio from playing

---

### (deck viewer): lazy loading on click, audio stops when playing a different one
https://user-images.githubusercontent.com/29434693/188330736-ca7af785-dcbe-4ce1-8779-223dd5170a5b.mp4

### (study page): fluid pause and resuming
https://user-images.githubusercontent.com/29434693/188330754-fb327439-11c3-43ae-b9d4-476371e8591f.mp4

### DIFF
--- a/src/models/lazy-audio.ts
+++ b/src/models/lazy-audio.ts
@@ -1,0 +1,114 @@
+/**
+ * Wrapper for the Audio object (HTMLAudioElement) that supports lazy loading/playing.
+ */
+export class LazyAudio {
+  private readonly audio: HTMLAudioElement = new Audio();
+
+  // whether or not the audio has enough data (readyState: 4 ==> HAVE_ENOUGH_DATA)
+  // we need to store this as a state because rewinding it resets this despite it being loaded
+  private hasEnoughData = false;
+
+  // whether or not the audio is already loading; prevents multiple network calls on double clicks.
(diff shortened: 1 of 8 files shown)


src/models/lazy-audio.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * Wrapper for the Audio object (HTMLAudioElement) that supports lazy loading/playing.
+ *
+ * Future: Is it better for this class to compose an Audio, or directly inherit from it?
+ * Compose Audio (Pros):
```
derricklee43, line 5: If we think we won't need to migrate to howler.js in the future, I think either is fine. But if so, I think we should be composed of an audio file

derricklee43 (approved):
These changes look good to me!

PR #67, opened 2022-09-07
feat: cancel ongoing fetches on unmount

jiejasonliu:
relatively simple; `fetch` supports passing in a signal in which you can abort to cancel the fetch


derricklee43 (approved):
also lgtm!

PR #70, opened 2022-09-08
feat(css): improve landing page on smaller devices

jiejasonliu:
### note to self: change base to `main` and rebase after `add-authorized-routes` merges
---

- landing page has improved styling for smaller screens (respect to height)
- header no longer breaks out when width < 500px
- center 404 page text


derricklee43 (approved):
lgtm

PR #71, opened 2022-09-09
fix: remove usage of stateful jwt value causing stale closure

jiejasonliu:
### i'm praying that it works this time
---

**tl;dr** stale closures resulted in outdated values for the auth token after refresh

---

Using `useState` gives the state and its updater, if we pass the state to an async function / callback such that a closure is created, the state is captured once. 

Any further updates on it will refer to the now stale value. 

Furthermore, `setState` is async in nature, so there's no guarantee to when the state value actually updates.

A common solution to this is to use `useRef` which is reference-based, but afaik this isn't an easy drop in solution for Recoil.

Instead, I opted to drop the usage of stateful values in the fetch wrapper:
- it's suppose to pure/stateless, we shouldn't have any side effects
- grabbing the value directly from local storage makes more semantic sense

### DIFF
--- a/src/hooks/api/use-fetch-wrapper.ts
+++ b/src/hooks/api/use-fetch-wrapper.ts
@@ -1,10 +1,10 @@
 import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
-import { useRecoilState } from 'recoil';
+import { useSetRecoilState } from 'recoil';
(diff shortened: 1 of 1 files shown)


derricklee43:
looks good to me! This JWT is no fun, lol

derricklee43 (approved):

PR #72, opened 2022-09-09
Fix use play lesson

derricklee43:
This took way longer than it should have, lol.

## Summary of Changes ##
- Fixed usePlayLesson hook
- Created played card history
- Added skip functionality 
- Added Replay functionality

https://user-images.githubusercontent.com/39753553/189438846-d828667b-f1af-450f-be37-1329ce2b0825.mov


src/hooks/use-play-lesson.ts
```diff
@@ -1,46 +1,127 @@
...
   }
 
-  function pauseLesson() {
-    pause();
+  function nextCard(
+    currentCard: LessonCard,
+    upcomingCards: LessonCard[],
+    completedCards: LessonCard[]
+  ) {
+    const newcompletedCards = [currentCard, ...completedCards];
```
jiejasonliu, line 84: nit: camel case

src/hooks/use-play-lesson.ts
```diff
@@ -1,46 +1,127 @@
...
+    const newCurrentCard: LessonCard = upcomingCards[0] ?? currentCard;
+    const newUpcomingCards = upcomingCards.length === 0 ? upcomingCards : upcomingCards.slice(1);
```
jiejasonliu, line 86: `slice` has a lot of case handling internally so this is equivalent to just `upcomingCards.slice(1)`

from the standard: 
> If start is greater than the index range of the sequence, an empty array is returned.

src/models/lazy-audio.ts
```diff
@@ -14,10 +14,16 @@ export class LazyAudio {
...
 
   public async play(): Promise<void> {
-    return this.loadAudio({
-      onPlayable: () => {
-        this.audio.play();
-      },
+    return new Promise<void>((resolve, reject) => {
+      this.loadAudio({
+        onPlayable: () => {
+          this.audio.addEventListener('ended', () => resolve(), { once: true });
```
jiejasonliu, line 20: can you clarify why this is needed?

jiejasonliu (commented):
😍

src/models/lazy-audio.ts
derricklee43: lol, that was a whoopsies. I was playing with something that didn't work out and I forgot to delete it. 

Good catch, lol

src/hooks/use-play-lesson.ts
derricklee43: oh dope! it should be fixed now

src/hooks/use-play-lesson.ts
derricklee43: lol, that was my bad trying to rename things. fixed

jiejasonliu (approved):
lgtm, nice work.
fyi cd doesn't work rn bc we are out of memory 🙃

PR #74, opened 2022-09-12
Add speech recognition

derricklee43:
## Summary of Changes ##
- Added Speech Recognition Hook
  - Works in all currently defined langauges
  - Added language codes for each defined language
- Connected Hook to study page
- Added pulsing icon when recording 
- Logs result to the console and records it on card
## TODO ##
(Will add tickets)
- Toggle Speech Recognition
- Toggle Animation
- Toggle Sound Effect
- Set Card Pause lengths 
- Add correct and incorrect animation and sound effect
- Add add additional sound effects
  - Finish
  - New card 
- Advanced settings menu (kebab or gear) (UI)
  - Toggle Animation
  - Toggle Sound Effects
  - Toggle Speech Recognition
  - Set Card pause lengths
- Improve String matching
  - Normalize speech recognition output
    - Homophones in English
    - Kanji to Kana in Japanese  
  - use an algorithm to determine  closeness 
- Fix speech recognition error when started up multiple times through spamming control buttons

## Demo ##
- My speaker only picked up my voice and not the card audio

https://user-images.githubusercontent.com/39753553/189710159-8a8c4ff2-68a7-4674-9ecc-ce6d6292fdc4.mov


src/hooks/use-play-card-audio.ts
```diff
@@ -63,14 +69,30 @@ export function usePlayCardAudio() {
     setActiveCardSide('front');
     await playCardAudio(frontAudio);
 
+    // Start capturing speech
+    const capturedSpeechPromise = captureSpeech(lessonCard.card.back.language as any);
```
jiejasonliu, line 73: if a card has a language of `Default`, it needs to inherit from the parent

couple ways to do this:
- hold a reference to the deck in card/lesson-card
- hold the deck's language in the card/lesson-card

src/hooks/use-play-card-audio.ts
```diff
@@ -63,14 +69,30 @@ export function usePlayCardAudio() {
...
     // wait before flip
     const backDuration = await backAudio.getDuration();
     await wait(getPauseLength(backDuration));
 
+    // Stop capturing speech if not already ended
+    stopCapturingSpeech();
+
+    // Currently if the user pauses while the recording is going, we will only get the results back before the pause.
+    // I think this is reasonable for now, but we might want to come up with a better system in the future.
+    const capturedSpeech = await capturedSpeechPromise;
```
jiejasonliu, line 84: what would the ideal behavior be?

src/hooks/use-speech-recognition.ts
```diff
@@ -0,0 +1,61 @@
...
+      speechRecognitionRef.current.onresult = (event) => {
+        const resultsLength = event.results.length - 1;
+        const ArrayLength = event.results[resultsLength].length - 1;
+        const { transcript, confidence } = event.results[resultsLength][ArrayLength];
+        resolve({ transcript, confidence });
+      };
+
+      // Always fires when disconnected
+      // TODO: Find a guarantee that it will always fire after onresult
+      // Or find another to handle resolving the promise if the user has not spoken
```
jiejasonliu, line 28: suggestion: we could pass in some type of timeout which is composed of `card audio length` + `the pause length` and resolve if it exceeds that.

src/hooks/use-speech-recognition.ts
```diff
@@ -0,0 +1,61 @@
...
+import { DeckLanguage, getLanguageCode } from '../models/language';
+
+export function useCaptureSpeech() {
+  const speechRecognitionRef = useRef(getSpeechRecognitionObj());
+  const [isCapturingSpeech, setIsCapturingSpeech] = useState(false);
+
+  return { captureSpeech, stopCapturingSpeech, abortSpeechCapture, isCapturingSpeech };
+
+  function captureSpeech(language: DeckLanguage) {
+    speechRecognitionRef.current.lang = getLanguageCode(language);
```
jiejasonliu, line 12: should we have a guard for `speechRecognitionRef.current`?
technically it will be `undefined` for unsupported browsers right?

src/hooks/use-play-card-audio.ts
```diff
@@ -42,6 +47,7 @@ export function usePlayCardAudio() {
   }
 
   function clearAudio() {
+    abortSpeechCapture();
     if (!activeAudioRef.current) {
       clearTimer();
```
jiejasonliu, line 52: also a bug(?) i observed, if you pause during a wait period and go to a previous card, it stays flipped on the 'back'.

during the `wait2` period: `playing front|wait1|playing back|wait2`, i think going previous should flip the card to the front, could add something like this before we return here:
```ts
clearTimer();
if (activeCard && activeCardSide === 'back') {
    setActiveCardSide('front');
}
return;
```

src/hooks/use-speech-recognition.ts
jiejasonliu: hmm nvm, i guess manually calling `stopCapturingSpeech` after the wait is basically doing the same thing

src/pages/study-page/study-page.scss
```diff
@@ -3,9 +3,27 @@
 
 .study-page {
   .study-page-content {
+    width: 500px;
```
jiejasonliu, line 6: what's this do? if it's something w the responsiveness breaking at <500px, I believe rebasing on main should fix it

src/hooks/use-play-card-audio.ts
derricklee43: Lol, forgot to do that. I should really leave some todo's lying around. 

I ended up restricting the language type when the lesson card is created

src/hooks/use-speech-recognition.ts
derricklee43: Yeah, this one is tricky. It seems to be working now, but Ideally we would want some guarentees

src/hooks/use-speech-recognition.ts
derricklee43: good catch. I fixed the typing on speechRecognitionRef.current and added a guard

src/hooks/use-play-card-audio.ts
derricklee43: hmm, yeah there seem to be some bugs with playback. I added the snippet but was unable to test it because of the updated audio. I'll test it some more once we get it all merged.

src/pages/study-page/study-page.scss
derricklee43: It was for styling the icon, but I fixed it so it is bound to the flashcard below.

src/hooks/use-play-card-audio.ts
derricklee43: hmmm, after thinking about it, this might be best. 

I was thinking we could restart the card if they paused during speech recognition, but that creates inconsistent behavior with the pause and play buttons. 

I guess we could restart the listening, then combine the results, but there might be some serious challenges with that.

jiejasonliu (approved):

src/models/lesson-card.ts
```diff
@@ -1,13 +1,36 @@
...
+}
+
+export function createNewLessonCard(
+  card: Card,
+  deck: Deck,
+  repeatDefinitionCount: number
+): LessonCard {
+  const { front, back, ...rest } = card;
+  const newFront = createNewLessonCardContent(front, deck.metaData.frontLang);
+  const newBack = createNewLessonCardContent(front, deck.metaData.backLang);
```
jiejasonliu, line 26: oof `front` should be `back` here

PR #75, opened 2022-09-13
feat: implement spaced repetition using the leitner system

jiejasonliu:
## Summary
- add workspace config for using `jest-runner` extension (codelens for tests)
- update card client (and interface) to deserialize dates from response
- implement spaced repetition using the leitner system

## Leitner System
- **Implementation:**
  - with a 5-box system (this can be configured relatively easily)
  - interface allows requesting `N` number of cards from a deck for studying
- **Boxes:**
  - the first box (BOX1) is a special box, it is always 'stale'
  - each box is a 'tier' of some sort, with a configured date of when a card is marked stale
    - (e.g.) cards in the last box (BOX5) are not stale until 2 weeks have occurred since the last study 
- **Scoring:**
  - all cards initially have a score of 0, hence in BOX1.
  - scoring a card correctly will increment the score, moving it up a box
  - scoring incorrectly will move the card back to BOX1 regardless of its correct position
- **Pooling (staleness):**
  - all cards that are stale are pooled (all BOX1 cards, any that are past due) to be selected from
  - if more cards are needed, remaining cards (unstale) are ranked by soonest to stale and chosen randomly

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,11 @@
 {
-    "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
-    },
-    "editor.tabSize": 2,
-    "editor.formatOnSave": true
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  },
+  "editor.tabSize": 2,
+  "editor.formatOnSave": true,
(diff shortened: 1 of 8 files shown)


derricklee43 (approved):
This looks good to me. It looks really clean!

PR #77, opened 2022-09-15
feat: add user registration form validation

jiejasonliu:
## Summary
- sign-up page reorganized form inputs
- sign-up page has form validation
  - only simple validation on client side
  - otherwise, complex ones handled on server (maybe client side should be more strict?)
- sign-in page works
- guard the registration pages if you are already signed in

## Preview

https://user-images.githubusercontent.com/29434693/190529319-ea9eef24-ac94-4d42-8c8a-bc997acedf3c.mp4

### DIFF
--- a/src/assets/_colors.scss
+++ b/src/assets/_colors.scss
@@ -14,3 +14,5 @@ $blue: #25283d;
 $blue-hover: #1e1f31;
 $light-placeholder: #636b8b;
 $dark-placeholder: #595c80;
+
+$light-red: #e16060;
--- a/src/hooks/api/use-user-client.ts
+++ b/src/hooks/api/use-user-client.ts
@@ -76,7 +76,7 @@ export function useUserClient() {
    * @side_effect Writes the auth JWT to local storage and recoil atom state
    * @returns True if no user credentials were correct and the side-effect occurred, otherwise false.
    */
-  async function login(username: string, password: string): Promise<boolean> {
+  async function login(email: string, password: string): Promise<boolean> {
     const numRetries = 0; // don't retry (e.g. 401 means incorrect user creds)
 
     // if jwt exists, we just refresh the token and early exit
@@ -90,11 +90,10 @@ export function useUserClient() {
     // authenticate the user
     try {
       const payload = {
-        username: username,
+        username: email,
         password: password,
       };
       const jwtData = await fetchWrapper.post('/Authenticate', payload, numRetries);
-      console.log(payload, jwtData);
       const authJwt = jsonToAuthJwt(jwtData);
       simpleLocalStorage.upsert(LocalStorageKeys.authJwt, authJwt);
       setAuthJwt(authJwt);
--- a/src/hooks/api/use-user-client.ts
+++ b/src/hooks/api/use-user-client.ts
@@ -94,6 +94,7 @@ export function useUserClient() {
         password: password,
       };
       const jwtData = await fetchWrapper.post('/Authenticate', payload, numRetries);
+      console.log(payload, jwtData);
       const authJwt = jsonToAuthJwt(jwtData);
       simpleLocalStorage.upsert(LocalStorageKeys.authJwt, authJwt);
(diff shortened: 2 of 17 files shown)


derricklee43:
damn, Fort Knox over here :laughing:

src/pages/sign-up/sign-up-page.tsx
```diff
@@ -2,47 +2,151 @@ import React, { useState } from 'react';
...
+    if (!data) {
+      throw new Error('An unknown error occurred while submitting the registration form');
+    }
+
+    if (data.statusCode === 200) {
+      const { id } = data.response;
+      console.log('Successfully registered:', id);
+      navigate(paths.signIn);
+      // TODO: redirect to sign up success page with button to login
+      // I don't think we should automatically log users in
```
derricklee43, line 119: Is there a reason why we don't let them in immediately after creating an account?

src/assets/_colors.scss
```diff
@@ -14,3 +14,5 @@ $blue: #25283d;
 $blue-hover: #1e1f31;
 $light-placeholder: #636b8b;
 $dark-placeholder: #595c80;
+
+$light-red: #e16060;
```
derricklee43, line 18: That's a good red:joy:

src/hooks/api/use-user-client.ts
```diff
@@ -20,11 +27,51 @@ export function useUserClient() {
...
+      const payload = registerUserInfoToJson(userInfo);
+      const response = await fetchWrapper.post('/Register', payload, numRetries);
+      if (isRegisterSuccess(response)) {
+        return { statusCode: 200, response };
+      } else {
+        throw new Error('Expected register success to have id (userId), but got none');
+      }
+    } catch (error) {
+      // data is an array of IdentityErrors
+      debugger;
```
derricklee43, line 64: Did you mean to keep this in?

src/pages/sign-in-page/sign-in-page.tsx
```diff
@@ -29,7 +31,17 @@ export const SignInPage = () => {
   );
 
   function handleSubmitClick() {
-    console.log(userName, password);
+    userClient.login(userName, password).then((success) => {
```
derricklee43, line 34: Is there a reason here for using `.then` instead of `await`?

derricklee43 (approved):
I like it! LGMT!

src/pages/sign-up/sign-up-page.tsx
jiejasonliu: i think this applies more to apps that have you verify an email + ability to change password

it's usually a good idea to let users try to login but it's different here i guess
no preference really so i'll change it to just log them in

src/hooks/api/use-user-client.ts
jiejasonliu: of course :trollface:

src/pages/sign-up/sign-up-page.tsx
derricklee43: oh, I see. I'm down with keeping it as is. Either way

src/pages/sign-in-page/sign-in-page.tsx
jiejasonliu: no, making the function async and awaiting this is equivalent
but i'll change it to be consistent w our other async handlers

src/pages/sign-up/sign-up-page.tsx
```diff
@@ -2,47 +2,151 @@ import React, { useState } from 'react';
...
+          newFormError.email = desc;
+        } else if (code.includes('Password')) {
+          newFormError.password = desc;
+        }
+      });
+      setFormError(newFormError);
+    }
+  }
+
+  function _getLabelError(
```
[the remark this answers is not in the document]
derricklee43: oh, I see. I like that. I think I'll start doing that for helpers

derricklee43:
lgtm

PR #79, opened 2022-09-19
added lesson results page

derricklee43:
## Summary of Changes ##
- Split the read-only flashcard variant into a separate component 
  - the flashcard component was already getting a bit complicated 
  - it made the typing for both components a lot tighter
  - read-only flashcard now has three colors/variants: `light-blue` | `red` | `green`
- StudyPage now has two subpages
  - StudyLessonPage
    - Contains all functionality originally in the study page
    - Added a stopwatch to record lesson length 
  - StudyLessonResultsPage  
    - I didn't want to send all the lesson cards through the URL 
      - However, if you think that would be better, I can switch it to that 
    - Shows all studied cards marked according to their outcome during the lesson
 - When Lesson finishes, a callback is called in StudyPage which switches out the page
 - Added a shitton of icons
 - Removed read-only-flashcard-set
   - extracted lazy audio player into a hook
   - it wasn't doing much, and I didn't want to have to complicate it with the colored/tagged studied flashcards 
   - moved tests around, and got rid of some that didn't make sense anymore

 

https://user-images.githubusercontent.com/39753553/191136609-8fed68fa-cd2e-409e-82c7-f827587b0448.mov


src/hooks/tests/use-stop-watch.test.ts
```diff
@@ -0,0 +1,121 @@
...
+import { useStopWatch } from '../use-stop-watch';
+
+describe('useStopWatch', () => {
+  const setup = () => {
+    const { result } = renderHook(() => useStopWatch());
+    return result.current;
+  };
+
+  it('should getElapsedTime after starting', () => {
+    withFakeTimers(async () => {
```
jiejasonliu, line 12: you'll need a return here since `withFakeTimers` is asynchronous (implied bc the inner callback is async)
alternatively, you can promote the outer scope to async and await it but the former (just returning) is better and equivalent

you need to do this bc jest awaits each test, and if you're not returning the promise, the test finishes before the promise resolves which leaks into other tests

the same applies to all tests in this file as well as `use-timer.test.ts`, not sure why I didn't catch that oops

src/pages/study-page/study-results-page/study-results-page.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <div key={card.key} className="study-results-flashcard">
+          <div className="study-results-flashcard-icon">{icon}</div>
+          <ReadOnlyFlashcard
+            variant={variant}
+            frontText={card.front.text}
+            backText={card.back.text}
+            onFrontSpeakerClick={() => playLazyAudio(card.front.audio)}
+            onBackSpeakerClick={() => playLazyAudio(card.back.audio)}
+          />
+          <div className="study-results-flashcard-icon"></div>
```
jiejasonliu, line 124: extra div?

jiejasonliu (approved):
really awesome PR, left some discussion comments and a requested test fix
also some nits but none of these are too major

architecture/flow-wise, this was great -- i found it super easy to follow!

src/pages/study-page/study-results-page/study-results-page.tsx
derricklee43: I was using this for spacing. Somehow I spaced using a variable and using a right-margin:joy:

src/hooks/tests/use-stop-watch.test.ts
derricklee43: huh. had no clue, lol. But that makes sense. It should be fixed now.

src/helpers/time.ts
```diff
@@ -0,0 +1,56 @@
...
+    date.getDate(),
+    date.getHours(),
+    date.getMinutes()
+  );
+}
+
+/**
+ * @returns returns the date formatted as MM/DD/YYYY
+ */
+export function getFormattedDate(date: Date) {
```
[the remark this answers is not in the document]
derricklee43: oh, that is way nice, haha.

src/components/deck-cover/deck-cover.tsx
```diff
@@ -49,6 +49,10 @@ export const DeckCover = ({
   }
 
   function getCoverBack() {
+    // TODO: We might have some reworking to do.
+    // I guessing we do not want to grab all the cards for every deck just
+    // calculate the percentStudied
+    // Maybe we should add a property to the deck
```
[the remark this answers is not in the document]
derricklee43: Yeah, getting the value right before sending makes a lot of sense

jiejasonliu (approved):
reapproved ✅ 
thanks for making these changes so quickly

PR #84, opened 2022-09-23
Added kebab menu to override outcome in Lesson Results Page

derricklee43:
## Summary of Changes##
- Added roles to `correct`, `incorrect`, and `skipped` icons for accessibility and testing
- Set `DropDownOption` generic to be on the `id` instead of `value`
- Added kebab menu to study results cards to override card outcome
  - kinda plain, might polish in future
-  broke study-results-cards into a separate component
- Readonly cards now break into two lines for small screens
  - Added index on view deck page cards to help tie card faces together  

https://user-images.githubusercontent.com/39753553/191891884-f35477aa-8d05-4632-8070-91b4c7f9d2f4.mov

### DIFF
--- a/src/components/read-only-flashcard/read-only-flashcard.scss
+++ b/src/components/read-only-flashcard/read-only-flashcard.scss
@@ -9,6 +9,12 @@ $card-height: 175px;
   width: 100%;
   gap: 20px;
   margin: 10px 0px;
+
+  @media screen and (max-width: $medium-screen) {
+    display: block;
+    margin: 25px 0px;
+  }
+
   .rof-card-face {
     display: flex;
     flex-grow: 1;
@@ -19,6 +25,12 @@ $card-height: 175px;
     height: $card-height;
     text-align: center;
     overflow-y: auto;
+
+    @media screen and (max-width: $medium-screen) {
+      &:first-child {
+        margin-bottom: 10px;
+      }
+    }
 
     .rof-button-strip {
       position: sticky;
--- a/src/components/search-bar/search-bar.tsx
+++ b/src/components/search-bar/search-bar.tsx
@@ -102,11 +102,11 @@ export const SearchBar = ({
   }
 
   function optionIncludesValueFilter(option: DropDownOption<string>) {
-    const optionVal = option.value;
+    const optionId = option.id;
     const searchVal = value.trim();
     return dropDownIgnoreCase
-      ? includesIgnoreCase(optionVal, searchVal)
-      : optionVal.includes(searchVal);
+      ? includesIgnoreCase(optionId, searchVal)
+      : optionId.includes(searchVal);
   }
 
   function hasText() {
--- a/src/components/search-bar/search-bar.tsx
+++ b/src/components/search-bar/search-bar.tsx
@@ -102,11 +102,11 @@ export const SearchBar = ({
   }
 
   function optionIncludesValueFilter(option: DropDownOption<string>) {
-    const optionId = option.id;
+    const optionVal = option.value as string;
     const searchVal = value.trim();
     return dropDownIgnoreCase
-      ? includesIgnoreCase(optionId, searchVal)
-      : optionId.includes(searchVal);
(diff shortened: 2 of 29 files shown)


src/components/read-only-flashcard/read-only-flashcard.scss
```diff
@@ -9,6 +9,12 @@ $card-height: 175px;
   width: 100%;
   gap: 20px;
   margin: 10px 0px;
+
+  @media screen and (max-width: $medium-screen) {
```
jiejasonliu, line 13: nice! but I still think it manages to get too small, perhaps we should use `$large-screen` (or even something in between lg/giant, idk)

src/components/search-bar/search-bar.tsx
```diff
@@ -102,7 +102,7 @@ export const SearchBar = ({
   }
 
   function optionIncludesValueFilter(option: DropDownOption<string>) {
-    const optionVal = option.value;
+    const optionVal = option.value as string;
```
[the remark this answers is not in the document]
derricklee43: I am using the generic to limit the id type even further. 
When this is used as the set of id's
```
const languages = [a, b, c] as const;
```
option.id is typed to be `a  |  b |  c` instead of `string`

But I agree that we should limit the id to extend `string` instead of `React.Key`, I forget we can further limit the value used for the react key. 

Would we want two generics on the `DropDownOption` component?

src/components/read-only-flashcard/read-only-flashcard.scss
derricklee43: yeah, I agree it gets too small. 

What do you think about breaking the navbar earlier? 

I've noticed when I have my phone horizontal, the sidebar doesn't break and takes up about a third of the screen. 

### Right before the nav breaks when nav breaks on large and cards break on medium 

Screen Shot 2022-09-23 at 11 43 44 AM

### Right before the cards break when nav breaks on large and cards break on medium 

Screen Shot 2022-09-23 at 11 43 29 AM

jiejasonliu: so like this? not a bad idea at all.

```diff
 .sidebar-layout-sidebar {
-  @media screen and (max-width: $large-screen) {
+  @media screen and (max-width: $giant-screen) {
     padding-left: unset;
     width: $sidebar-width / 1.15;
   }

-  @media screen and (max-width: $medium-screen) {
+  @media screen and (max-width: $large-screen) {
     display: none;
   }
 }
@@ -47,11 +47,11 @@ $content-height: calc(100vh - #{$header-height + $content-padding});
   margin-left: $sidebar-width;
   padding-top: $header-height;

-  @media screen and (max-width: $large-screen) {
+  @media screen and (max-width: $giant-screen) {
     margin-left: ($sidebar-width / 1.15);
   }

-  @media screen and (max-width: $medium-screen) {
+  @media screen and (max-width: $large-screen) {
     margin-left: 0;
   }
```

we would need to change breakpoints for the navigation hamburger `.c-header-content` and some other things if we're hiding the sidebar early though

src/components/search-bar/search-bar.tsx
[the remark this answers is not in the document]
jiejasonliu: yeah, that makes sense -- two generics to filter down the type.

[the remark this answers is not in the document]
derricklee43: Cool. I just added it

src/components/read-only-flashcard/read-only-flashcard.scss
derricklee43: Ok, I think I got it all updated. How does this look?

https://user-images.githubusercontent.com/39753553/192109663-7ea3131f-57a4-4bc8-8bfe-31dcb69f3486.mov

jiejasonliu: this looks great!

jiejasonliu (approved):

PR #85, opened 2022-09-23
feat: add account popup

jiejasonliu:
## summary!
- adds the account popup
  - can close when outside click or escape press
- updates header to show responsive account group 
- account pfp from Gravatar (defaults to unique 8-bit icon like github if not registered)
- new recoil selector (userInfoState) that has a dependency on the authJwt atom
  - whenever the jwt updates, it asynchronously fetches the user info
  - this is why the `log out` button is seen for a second, it's the `React.Suspense` fallback state

---

https://user-images.githubusercontent.com/29434693/191907829-11321d98-2520-4c0b-af57-af6ca57a5e62.mp4

### DIFF
--- a/src/hooks/use-key-press.ts
+++ b/src/hooks/use-key-press.ts
@@ -17,17 +17,17 @@ export const useKeyPress = (
   useEffect(() => {
     if (!active) return;
 
-    const handleEscapeKey = (event: KeyboardEvent) => {
+    const handleKeyPress = (event: KeyboardEvent) => {
       if (event.key === key) {
         action(event);
       }
     };
-    document.addEventListener('keydown', handleEscapeKey);
+    document.addEventListener('keydown', handleKeyPress);
 
     // destructor; called when component using this hook gets destroyed or a dependency changes
     return () => {
       if (!active) return;
-      document.removeEventListener('keydown', handleEscapeKey);
+      document.removeEventListener('keydown', handleKeyPress);
     };
   }, [key, action, active, ...extraDeps]);
 };
--- a/src/hooks/use-key-press.ts
+++ b/src/hooks/use-key-press.ts
@@ -1,0 +1,44 @@
+import { useEffect } from 'react';
+
+/**
+ * Perform an action (callback) whenever a key press occurs.
+ *
+ * @param key the key string pressed (e.g. 'Tab' or 'Escape')
+ * @param action callback handler after a click outside occurs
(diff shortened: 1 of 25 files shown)


src/state/auth-jwt.ts
```diff
@@ -11,6 +17,45 @@ export const authJwtState = atom<AuthJwt | undefined>({
...
+  key: 'userInfoStateAsync',
+  get: async ({ get }) => {
+    const authJwt = get(authJwtState);
+
+    if (!isAuthJwt(authJwt)) {
+      return undefined;
+    }
+
+    // we unfortuantely cannot use `useFetchWrapper` since that adds hooks (breaks rule of hooks)
+    // either we replace this selector with effects in a nested component that updates this atom
```
jiejasonliu, line 40: i've seen people put a `RecoilEffectListener` component as a sibling of `App` and then that file basically has a component that has an effect that sets up the dep chain.

(e.g. if atomB's result depends on atomA)
```ts
const fetchWrapper = useFetchWrapper();

useEffect(() => {
   (async () => {
     const someResult = await fetchWrapper.get(...);
     setAtomB(someResult);
   })();
}, [atomA])
```

that way, we could use any hook we want bc it lives in a component.
I won't be doing this atm but if growing pains occur due to async selectors/atoms, we can move to this pattern

src/hooks/use-key-press.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect } from 'react';
```
derricklee43, line 1: This is really dope!! Ima have to add this to the study lesson, haha

src/state/auth-jwt.ts
derricklee43: That makes sense :thumbsup:

derricklee43: That's actually pretty cool

derricklee43 (approved):
LGMT! That dropdown is so slick

src/hooks/use-key-press.ts
jiejasonliu: that would be so cool. it's pretty barebones right now, so depending on what you need, maybe allow it to take an array of keys or an object of keys to callbacks.

derricklee43: Yeah, an object of keys to callbacks sounds ideal. We can always figure it out later, haha

PR #87, opened 2022-09-24
fix: add promise lock to requests during token refresh

jiejasonliu:
## note to self: rebase on main

fixes a race condition between multiple refresh requests.
if the access token is expired and 2+ authorized requests are made at nearly the same time, all of them would try to refresh.

only one request at most should be refreshing and other ones (which got a 401) should be waiting until it completes

### solution
- add a promise lock (mutex) to the refresh
- if another requests tries to refresh but one is happening already:
  - skip the refresh
  - retry the original request when the promise lock resolves (i.e. refresh finishes)

### DIFF
--- a/src/helpers/promise.ts
+++ b/src/helpers/promise.ts
@@ -1,0 +1,23 @@
+/**
+ * Creates a deferred promise.
+ * Use sparingly: https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern
+ * The advantage of this is the ability to resolve a promise outside of its Promise scope.
+ */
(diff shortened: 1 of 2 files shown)


derricklee43 (approved):
This'll be so nice, haha. LGTM!

jiejasonliu:
yes! unsure if you experienced it, but sometimes you just get logged out randomly. 

it redirects you to login if refresh fails, which is guaranteed to occur if two refreshes happen with the same token pair.

so this should fix that observed issue.

derricklee43:
Haha, yeah that happened to me too. Dope!!

PR #88, opened 2022-09-24
added lesson sound effects

derricklee43:
https://user-images.githubusercontent.com/39753553/192110997-2cd30b85-6c33-4579-965c-66916468a838.mov


jiejasonliu (approved):
LGTM.

fyi lazy audio, audio chaining is something we need to revisit later.

it doesn't work on iOS for some reason, I narrowed it down to a `NotAllowedError` on `.play()`, it's swallowing perms during promise chain for some reason.

i've read some areas where audio files are preloaded (call load?) but then paused immediately so that it can be resumed without permission errors later.

but we can tackle this later, and potentially together since i have no way of debugging it concretely on my iOS device without a Mac 😦

derricklee43:
Huh, yeah that is strange, but yeah! I'm down for a debug sesh!

PR #90, opened 2022-09-28
trim transcripts before comparison

derricklee43:


jiejasonliu (approved):
was the captured speech actually possible to have whitespace?
LGTM though, mergeable asap

derricklee43:
haha, now that you mention it, props not. I can remove that in a late PR

PR #92, opened 2022-09-28
fix: clear promise lock on resolve

jiejasonliu:
i hope this fixes the sign out problem

if it ever happens again, can you send me the console logs?


derricklee43 (approved):
lgtm!

PR #93, opened 2022-09-29
feat: use howler.js for LazyAudio

jiejasonliu:
drop in replacement with `howler.js` for `LazyAudio`
however, instead of composing `Howl`, we inherit from it

hopefully this allows mobile devices to autoplay after the first user interaction
if not, we can think of other solutions, but I think howler is worth to keep due to simplicity

### breaking changes
- cannot use addEventListener anymore, since `LazyAudio extends Howl`
  - either use `on` or `once` with howler's built in events


derricklee43 (approved):
LGTM!

PR #94, opened 2022-10-03
feat: update card scores when lesson is finished

jiejasonliu:
## studying
- utilize spaced repetition when picking cards
- lesson size updated to be minimum of cards or 5 -- temporary until lesson config
- scores are updated when the `lesson finished` button is clicked
- **[bug fix]** you can no longer skip more than there are cards available
- **[bug fix]** stop lingering audio when lesson finishes by skipping or when clicking browser tab back button

## misc
- clicking on registration buttons in the header no longer resets the tab's history stack 
- dates from the server are serialized as UTC time (with assumption that server's time-zone stays UTC)
-  **[bug fix]** ensure listeners are registered before `audio.load()` is called
    - resolves an issue where audio doesn't play if the audio file was too small
    - to be specific, `durationAsync()` was never resolving since it loaded before the listener was registered

### DIFF
--- a/src/pages/study-page/study-results-page/study-results-page.tsx
+++ b/src/pages/study-page/study-results-page/study-results-page.tsx
@@ -9,7 +9,9 @@ import { StarIcon } from '../../../assets/icons/star-con/star-icon';
 import { BubbleDivider } from '../../../components/bubble-divider/bubble-divider';
 import { Button } from '../../../components/button/button';
 import { getFormattedMilliseconds } from '../../../helpers/time';
+import { useCardsClient } from '../../../hooks/api/use-cards-client';
 import { usePrompt } from '../../../hooks/use-prompt';
+import { MAX_SCORE } from '../../../hooks/use-spaced-repetition';
 import { Deck } from '../../../models/deck';
 import { LessonCard } from '../../../models/lesson-card';
 import { paths } from '../../../routing/paths';
@@ -28,6 +30,7 @@ export const StudyResultsPage = ({
   lessonTime,
   onLessonCardsChange,
 }: StudyResultsPageProps) => {
+  const cardsClient = useCardsClient();
   const navigate = useNavigate();
 
   // block navigation if finish isn't clicked; otherwise, redirect back to the deck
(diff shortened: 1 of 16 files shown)


src/pages/study-page/study-results-page/study-results-page.tsx
```diff
@@ -92,4 +118,49 @@ export const StudyResultsPage = ({
...
+    // create promise to update cards with new scores
+    for (const lessonCard of lessonCards) {
+      if (lessonCard.outcome === 'unseen') {
+        continue;
+      }
+
+      // spaced repetition: if correct, score is increased by 1; otherwise reset to 0
+      let newScore = 0;
+      if (lessonCard.outcome === 'correct') {
+        newScore = Math.min(lessonCard.score + 1, MAX_SCORE);
```
derricklee43, line 154: Does the MAX_SCORE correspond with the number boxes in the spaced repetition system?

src/pages/study-page/study-results-page/study-results-page.tsx
```diff
@@ -92,4 +118,49 @@ export const StudyResultsPage = ({
...
+      }
+
+      if (lessonCard.id) {
+        promises.push(cardsClient.updateCardScoreById(lessonCard.id, newScore));
+      } else {
+        throw new Error(`Failed to update lesson card: id was undefined`); // this shouldn't be possible...
```
derricklee43, line 160: I am starting to think we should have two card types: one with the id required and the other with it optional. Then we would just use the optional type in the deck editor, so we can avoid all these checks.

derricklee43 (approved):
LGMT!

src/pages/study-page/study-results-page/study-results-page.tsx
jiejasonliu: it corresponds with the number of boxes - 1 since it follows 0-based indexing

src/pages/study-page/study-results-page/study-results-page.tsx
jiejasonliu: probably not apart of this PR but we could make `id` required on the current cards and then the new type would be something like this:

```ts
type NewCard = Omit<Card, 'id'> & Partial<Pick<Card, 'id'>>`
```
which will create a `Card` type with the `id` optional.

src/pages/study-page/study-results-page/study-results-page.tsx
derricklee43: oh, haha. Yeah, that makes sense!

PR #97, opened 2022-10-05
refactor: use path aliases for parent imports

jiejasonliu:
## note: rebase on main after `add-export-cards-popup` merges

reviewing is probably easiest with a commit-by-commit basis

purely a code style change, we now always prefer `@/...` over parent imports `../../...`
because honestly, the relative parent paths were getting way too long
* the way this works is that `@/...` is an alias for the absolute path `src/...`

this is also applied for our scss files, we won't have any ambiguity over the relative path,
we can just stick `@import '@/assets/styles'` and we're guaranteed for it to work

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,7 @@
     "extends": [
         "plugin:react/recommended",
         "plugin:@typescript-eslint/recommended",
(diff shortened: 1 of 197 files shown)


derricklee43 (approved):
This is sooo dope!! LGTM!!

PR #99, opened 2022-10-09
fix: improve endpoint robustness

jiejasonliu:
- prevent empty cards (both front & back empty) from being added/saved
- use the new batch endpoints for deck create, card edit/score update/create/delete
- prevent deck editor from sending empty requests (e.g. no cards deleted should avoid doing `Cards/Delete`)
- make account popup fixed when header is also fixed


jiejasonliu:
> Everything looks good to me! sorry it took so long to review😅

no worries, who works on break 😆 
also sorry for the merge conflicts in advance

PR #100, opened 2022-10-14
Add category search bar

derricklee43:
## Summary of Changes
- Added Category Search Bar
  - Regular Search bar also still exists
- Broke Dropdown Button into a separate component for reuse
  - There was no class prop to target the `DropdownMenu` button directly, so I didn't want to target through the structure
     - I might add a class prop there so that we can set it directly instead of having the duplicate code
-   Added `useSearchCategories` hook
    - Provides the default categories of personal decks, public decks, and users
    - Only personal decks are populated, public decks and user will be a separate MR 
    - Handles switching between categories and filtering results
- Renamed `userClient` to `accountClient`, so we add a separate `useClient` to fetch other users

https://user-images.githubusercontent.com/39753553/195956619-a8a37d63-4c13-4e53-8047-e159e57f088f.mov


src/components/search-bar/category-search-bar/search-bar-category-dropdown/search-bar-category-dropdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+  categories,
+  onCategorySelect,
+  onClick,
+}: SearchBarCategoryDropdownProps<T, S>) {
+  const [isOpen, setIsOpen] = useState(false);
+
+  const categoryDropdownRef = useRef(null);
+  useOutsideClick(categoryDropdownRef, () => setIsOpen(false));
+  useFocusTrap(categoryDropdownRef, isOpen);
+  useEscapePress(() => setIsOpen(false));
```
jiejasonliu, line 27: both `useOutsideClick` and `useEscapePress` should only be active when `isOpen` is true, so let's pass that in as the third and second arg respectively

src/components/header/header.tsx
```diff
@@ -33,9 +35,22 @@ export const Header = ({
 
   const [navToggled, setNavToggled] = useRecoilState(navToggledState);
 
-  const userClient = useUserClient();
+  const userClient = useAccountClient();
```
jiejasonliu, line 38: should we rename the variable to `accountClient` since we're adding a user client later?

src/hooks/use-search-result-filter.ts
```diff
@@ -0,0 +1,22 @@
...
+    return searchOptions.filter(({ value }) => resultContainsValue(value, searchValue));
+  }
+
+  function resultContainsValue(result: string, searchValue: string) {
+    result = normalizeString(result);
+    searchValue = normalizeString(searchValue);
+    return result.includes(searchValue);
+  }
+
+  function normalizeString(str: string) {
```
jiejasonliu, line 16: is this different from the `normalize` method you added in `string.ts`?

src/hooks/use-search-categories.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const promises = [];
+    if (decks.length === 0) {
+      promises.push(fetchDecksAndRefresh());
+    }
+    // TODO: Add the other categories
+    await Promise.all([promises]);
+    setIsLoading(false);
+  }
+
+  async function fetchDecksAndRefresh() {
```
jiejasonliu, line 61: if we load `/decks` then we would fetch twice right?
i guess it's not a problem but would it make sense to fetch **only** if `userDecks` (the recoil value) is undefined?

i'm not sure about any issues that would cause but it's something to think about

jiejasonliu (commented):
overall this is great, i'm super impressed with this!!
i left a few nits and some potential control flow discussions

src/components/drop-down/drop-down-button/drop-down-button.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../../assets/colors';
+@import '../../../assets/styles';
```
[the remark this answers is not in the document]
derricklee43: lol, I forgore about this after merge update in:sob::sob: Thanks for the callout

src/components/search-bar/category-search-bar/search-bar-category-dropdown/search-bar-category-dropdown.tsx
derricklee43: I ended up scrapping this file and adding a `buttonClassName` to the `Dropdown` because it is doing the exact same thing, and the dropdown has it right.

src/components/header/header.tsx
derricklee43: haha, good point. I'll get those fixed

src/hooks/use-search-result-filter.ts
derricklee43: no, it's the same. I deleted it here, but forgore to push:sob:

src/hooks/use-search-categories.tsx
derricklee43: I am checking if the `decks` is an empty array where I call this function, so I can conditionally add the promise to the list of fetch resource promises. 

The default value for `decks` is `[]`, I tried switching it to undefined, but I was running into problems. I was planning on fixing this, and shifting around the control flow a bit in a separate MR where I will also add fetching for users and public decks.

derricklee43:
> overall this is great, i'm super impressed with this!! i left a few nits and some potential control flow discussions
> 
> **edit:** happy 💯 pull request sheeesh

🥳 🥳

src/hooks/use-search-categories.tsx
jiejasonliu: you're right, this wouldn't work since we'd have no way of knowing if a request is already fired

it'd probably include something regarding exposing whether a request is already occurring and preventing it...

so let's just worry about this later if it ever comes up

jiejasonliu (approved):
LGTM with the new changes. good stuffff 😀

PR #102, opened 2022-10-16
Complete search page

derricklee43:
## summary of changes
- updated search page
- created deck tiles 
  - We can switch it to a `DeckCover` but we would have to alter the back side for public decks if you'd like
- created user tiles
- dropped `isLoading` field, now we tell if a resource is loading if the corresponding `searchResults` is undefined
- decks atom now defaults to undefined
- created users and publicDecks atoms as well
- on `SearchValue` change the resources of the selected category are checked if undefined, if undefined, a request is made to fetch them

  
https://user-images.githubusercontent.com/39753553/196042253-f925d327-43a5-4422-9034-e4a6a0f41821.mov


src/pages/search-page/search-page.tsx
```diff
@@ -26,19 +27,97 @@ export const SearchPage = () => {
...
+  }
+
+  function getUserTiles() {
+    const users = categorySearchResults['users'];
+    if (users === undefined) {
+      return;
+    }
+    return (
+      <div className="search-page-user-tiles">
+        {users.map((user) => (
```
derricklee43, line 91: TODO: Move map out and save into variable to match `getDeckTiles`

src/assets/_styles.scss
```diff
@@ -93,6 +93,42 @@ $z-popup-modal: 1024;
...
+  height: 100%;
+
+  @media screen and (max-width: $medium-screen) {
+    margin: $mobile-gutter-width 0px;
+    row-gap: $mobile-gutter-width;
+    column-gap: $mobile-gutter-width;
+  }
+}
+
+@mixin outlined-tile($width, $height) {
```
derricklee43, line 114: Move mixin back into `user-tile.scss`

src/components/deck-tile/deck-tile.scss
```diff
@@ -0,0 +1,60 @@
+@import '@/assets/styles';
+@import '@/assets/colors';
+
+.deck-tile {
+  background-color: transparent;
+  display: inline-flex;
+  // justify-content: start;
```
derricklee43, line 7: remove commnets

src/components/button/button.scss
```diff
@@ -14,43 +14,43 @@
...
-  &.light {
-    @include white-hover;
-    background-color: $white;
-    border-color: $blue;
-    color: $blue;
-
-    &:disabled {
-      border-color: $grey;
-      color: $grey;
+.dark-button {
```
jiejasonliu, line 23: changes here cause the selector specificity to change breaking a lot of hover effects across the app such as
  - account popup buttons
  - bubble dropdown
  - ...and probably more

i'm assuming there was a reason this was done so that custom styles worked on the search page tiles.
a solution could be to fix all those areas affected but i'm not sure about the breadth of it beyond those

perhaps it would be better to move these selectors back to the way they were (nested inside `.button`) and then you could be more specific in your custom styles (e.g. `.button.deck-tile` instead of `.deck-tile`) to get the styles you want

src/hooks/api/use-public-users-client.ts
```diff
@@ -0,0 +1,37 @@
...
+  // GET: /Users/Names
+  async function getPublicUsernames() {
+    const usernames: string[] = await fetchWrapper.get('/users/names');
+    return usernames;
+  }
+
+  // GET: /Users/Username
+  async function getPublicUsers() {
+    const usernames = await getPublicUsernames();
+    const allUsersRequests = usernames.map((username) => fetchWrapper.get(`/users/${username}`));
```
jiejasonliu, line 24: i feel like we should have an endpoint that gives all public users as well as some extra data like their `numDecks` so that would don't have to make API calls to the order of users

src/components/user-tile/user-tile.tsx
```diff
@@ -0,0 +1,22 @@
...
+  user: PublicUser;
+  onClick?: () => void;
+}
+
+export const UserTile = ({ user, onClick }: UserTileProps) => {
+  const userName = user.username;
+  const numDecks = user.publicDecks.length;
+  return (
+    <Button className="user-tile" onClick={() => onClick?.()} variant="invisible">
+      <ProfilePicture username={user.username} showGlow={false} className="user-tile-profile-pic" />
```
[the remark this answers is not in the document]
derricklee43: Yeah, that makes sense.  Should a user have access to the emails of all other users? 

Would be better of sending over a URL? I think that would also make adding a profile picture easier if we decide to implement that.

src/components/deck-tile/deck-tile.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  .dt-title {
+    color: $white;
+    font-size: 20px;
+    margin-bottom: 4px;
+  }
+
+  .dt-description {
+    flex-grow: 1;
+    overflow-y: scroll;
```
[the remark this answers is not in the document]
derricklee43: whoops. I switched it to `auto`, thanks!

src/pages/search-page/search-page.tsx
```diff
@@ -26,19 +26,77 @@ export const SearchPage = () => {
...
+    const showAuthor = category === 'public decks';
+    const deckTiles = decks.map((deck) => {
+      const author = showAuthor ? deck.metaData.ownerId : undefined; // TODO: Ask Mason or Jon to add the author email and replace
+      const id = deck.metaData.id.toString();
+      return (
+        <DeckTile
+          key={id}
+          title={deck.metaData.title}
+          description={deck.metaData.desc}
+          numCards={deck.cards.length}
```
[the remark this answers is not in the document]
derricklee43: yeah, I'll just wait for your changes to be merged into main, then pull them in and update it

src/components/user-tile/user-tile.tsx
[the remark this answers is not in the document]
jiejasonliu: we should probably have a discussion on that since we can get a user's public data via their username OR email
if not, we could also have the backend hash their email and send it instead

because the PFP URL is really just a static endpoint with the only changing data the md5-hashed email

jiejasonliu:
merge conflicts result from:
- flashcard-decks being **moved** to `pages/decks-page/...`
- profile-page return has a react fragment `<>` wrapping it; there are no other changes, this can be removed

jiejasonliu (approved):

PR #103, opened 2022-10-16
feat: add study configuration popup

jiejasonliu:
- add study config popup
  - two types: **study new cards** (all cards unsure what we mean by new), and **review cards** (via spaced repetition)
  - allow order cards from lesson if **study new cards** selected
  - set max cards with number validation
- serialize study config settings as a query param when navigation to study page
- applied to all study flows (decks page, view deck, profile page decks)
  - clicking 'study again' after a lesson will persist the query param 

#### note: deck progress bar is a dummy value until we have that figured out

---

https://user-images.githubusercontent.com/29434693/196024346-b799e961-e8b5-41e8-8823-b8ade9bce7d5.mp4


derricklee43 (approved):
LGTM! The popup looks dope!

derricklee43:
What do you think about moving the study config popup to the lesson page, and it opens up when first loaded? 

After thinking about it some more, it seems like it is a pain to split those two apart, and we will have to repeat it every time we add a link to the study page. 

But if you prefer it how it is now, I'm down with it!

jiejasonliu:
> What do you think about moving the study config popup to the lesson page, and it opens up when first loaded?
>
> After thinking about it some more, it seems like it is a pain to split those two apart, and we will have to repeat it every time we add a link to the study page.

good idea! i think that's reasonable because yeah this would be duplicated across every study flow, it shouldn't be too much extra work

jiejasonliu:
@derricklee43, i pushed a commit that makes the config popup load on the study page
- show popup if there are no query params
- start lesson button navigates to the same route with query params which starts the lesson
- if the popup is quit, it begins a default session which can be useful for debugging/opting out of config
- clicking `study again` after the lesson finishes will show the popup again but with the same settings prefilled

---


https://user-images.githubusercontent.com/29434693/196088278-3897c6c0-4564-4e5d-9029-c29234e825c9.mp4

derricklee43:
This is so dope! Everything looks good to me!

PR #104, opened 2022-10-18
feat: mobile accessibility improvements

jiejasonliu:
- close sidebar on content click/escape key when open
  - **caveat:** doesn't close on header click; possible, but would be a lot more code
- various spacing improvements
- view deck uses css grid to layout buttons on smaller dimensions

---

https://user-images.githubusercontent.com/29434693/196317578-cfe0ac6d-5e02-4aca-863f-59861f92f6b7.mp4


derricklee43:
LGTM! So down to not collapse sidebar on header click:laughing:

PR #107, opened 2022-10-24
fix: text overflow improvements and save button state

jiejasonliu:
### fetch wrapper
- fixes a bug introduced last PR where it wouldn't abort fetches on unmount if there hadn't been a refresh before
- i still can't reproduce the logging out bug, lmk if it occurs again and i'll look into it some more

### text overflow
- introduce `grid-center` which performs safe centering when the inner container overflows the parent
- introduce `multiline-ellipsis-overflow` which allows containers to describe the max lines before overflowing
- improve text overflow on these areas:
  - flashcard decks
  - view flashcard
  - study page popup, study page (+ lesson card), and results page

### deck editor
- disable save button when there are no changes
- fix race condition caused by two requests modifying a deck
  - can observe that sometimes when editing title+desc and adding cards, only the cards are updated

---

https://user-images.githubusercontent.com/29434693/197456560-6f5ff842-f15c-41da-abee-a60b36da5d90.mp4

https://user-images.githubusercontent.com/29434693/197456565-bb5e1d33-f64c-407f-94bd-ac1716134068.mp4

### DIFF
--- a/src/hooks/use-deck-editor.ts
+++ b/src/hooks/use-deck-editor.ts
@@ -68,14 +68,18 @@ export const useDeckEditor = (deck: Deck): DeckEditorReturn => {
       const updatedCards = Object.values(state.updatedCards).filter(filterBlankCards);
       const deletedCards = Object.values(state.deletedCards);
 
-      const promises = [
-        decksClient.updateDeckById(state.currentDeck),
+      // alone to prevent race condition on backend since updating cards mutates a deck as well
+      // this resolves the issue where updating both metadata and cards can cause the metadata to not save
+      await decksClient.updateDeckById(state.currentDeck);
+
+      const cardPromises = [
         ...[addedCards.length > 0 ? cardsClient.addCards(addedCards, deckId) : []],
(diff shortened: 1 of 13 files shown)


jiejasonliu:
@derricklee43 can you see if these style changes work on your mac machine? a lot of these are webkit specific fixes which `caniuse` reports as being supported for all browsers but i can't verify myself

src/hooks/use-deck-editor.ts
```diff
@@ -68,14 +68,18 @@ export const useDeckEditor = (deck: Deck): DeckEditorReturn => {
       const updatedCards = Object.values(state.updatedCards).filter(filterBlankCards);
       const deletedCards = Object.values(state.deletedCards);
 
-      const promises = [
-        decksClient.updateDeckById(state.currentDeck),
+      // alone to prevent race condition on backend since updating cards mutates a deck as well
```
jiejasonliu, line 71: actually gonna revert this and fix it on the backend side bc i've identified the cause now

derricklee43 (approved):
This looks good to me! I ran the branch on my machine and it looked good!

PR #108, opened 2022-10-25
Add public page

derricklee43:


src/components/resource-loader/resource-loader.tsx
```diff
@@ -58,3 +58,25 @@ export function loadDeck(deckId: string | undefined, allowUndefinedDeckId = fals
     return deck;
   };
 }
+
+export function loadPubicDeck(deckId?: string) {
+  // hooks outside of functional components must be consumed immediately by the component
+  const { getPublicDeckById } = useDecksClient();
+  const { getPublicCardsById } = useCardsClient(); // TODO: A big error happens when this useCardsClient hook is removed
```
jiejasonliu, line 65: the big error should only occur if adding/removing hooks during hot reload right?
i think during development `resource-loader` is difficult since its closure based, so modifying hook order makes it stale and throw that error

but once you're not touching it anymore, the order of hooks is deterministic and the error shouldn't occur

jiejasonliu: also nit: spelling on function name 😆

src/pages/profile-page/personal-profile-page/personal-profile-page.tsx
```diff
@@ -42,6 +45,7 @@ const AsyncPersonalProfilePage = () => {
         <PageHeader label="my profile" />
       </div>
       <UserDetails
+        profilePicUrl={profilePicUrl}
         username={userData?.username ?? ''}
         email={userData?.email}
         dateJoined={new Date()}
```
jiejasonliu, line 51: do we have this data?

jiejasonliu (approved):
love it! the bubble tags are such a nice touch.

do you think we would ever want to support different colors for different tags?
(i.e. able to mix and match colors instead of only one)

src/components/resource-loader/resource-loader.tsx
derricklee43: The error was happening when I was navigating from a public deck to the newly copied deck. I believe React doesn't see two separate `ResourceLoader` components, but one `ResourceLoader` with changing hooks.

It is fine now that the hooks are the same between the two helper functions, but if we ever need to add additional hooks in the future, we might run into issues.

derricklee43: > also nit: spelling on function name 😆

:joy: it's fixed now

src/pages/profile-page/personal-profile-page/personal-profile-page.tsx
derricklee43: It wasn't in the interface, so I wasn't sure if it existed yet. It's there; and I've fixed it

src/pages/profile-page/personal-profile-page/personal-profile-page.tsx
```diff
@@ -39,16 +39,21 @@ const AsyncPersonalProfilePage = () => {
     fetchDecksAndRefresh();
   }, []);
 
+  if (userData === undefined) {
+    return <LoadingPage />;
```
jiejasonliu, line 43: was this needed? does suspense not handle this for us?

src/pages/view-deck-pages/view-personal-deck-page/view-personal-deck-page.scss
```diff
@@ -10,8 +10,7 @@
     grid-auto-flow: row;
     grid-template-columns: auto min-content;
     grid-template-rows: min-content auto;
-    column-gap: 16px;
-    overflow-wrap: anywhere;
```
jiejasonliu: we probably want to keep this (merge conflict)

jiejasonliu:
alright, gave another review; looks good to merge after! 🚀

src/pages/view-deck-pages/view-personal-deck-page/view-personal-deck-page.scss
derricklee43: Oh, good catch; I'll get that updated.

src/pages/profile-page/personal-profile-page/personal-profile-page.tsx
jiejasonliu: oh I see. it's fine to keep, the only time that `userData` can be `undefined` is if the auth token expires in which case we shouldn't be able to hit this page anyhow

PR #110, opened 2022-10-27
connect-study-percents

derricklee43:
- I tried adding the studied/mastered progress bar, but I was having trouble with the styling and making it look good. 
  - I just punt it, and maybe look at it again in the future, lol
- added card count to deck tile
- added progress bar to view deck page 
Screen Shot 2022-10-26 at 9 45 03 PM
Screen Shot 2022-10-26 at 9 45 20 PM


src/pages/study-page/study-results-page/study-results-page.tsx
```diff
@@ -126,6 +126,16 @@ export const StudyResultsPage = ({
     );
   }
 
+  function getTotalDeckProgress() {
+    const numPointsEarned = lessonCards.filter((card) => card.outcome === 'correct').length;
```
derricklee43, line 130: I think this function might have too much context. Maybe we could make an API endpoint to get the right number

jiejasonliu: it depends, would it be an endpoint that calculates the percentage "if it was actually applied"?
i feel like that would be expensive since it would need to hit an endpoint each time we mark correct/incorrect.

now, what are we suppose to be showing? `numPointsEarned` should include the current deck no?

if this is based off the mastery (all cards having a score of 4), we should change it to say 'deck mastery percentage' and have a better algorithm for doing it (something involving `deck.cards`)

otherwise, if we just want to show deck study progress, it'd just be the number of non-zero scored cards (something like this, i slapped it together):
```ts
  function getTotalDeckProgress() {
    const lessonCardsById = new Map(lessonCards.map((card) => [card.id, card]));
    const newPoints = deck.cards
      .map((card) => {
        const lessonCard = lessonCardsById.get(card.id);
        if (lessonCard?.outcome === 'correct') {
          card.score += 1;
        } else if (lessonCard?.outcome === 'incorrect') {
          card.score = 0;
        }
        return card;
      })
      .filter((card) => card.score > 0).length;

    const totalPossiblePoints = deck.cards.length;
    const progress = (newPoints / totalPossiblePoints) * 100;

    if (progress === 0) {
      return '0%';
    }

    return progress < 1 ? '~1%' : `${Math.round(progress)}%`;
  }
```

src/pages/view-deck-pages/view-personal-deck-page/view-personal-deck-page.tsx
```diff
@@ -17,6 +18,8 @@ interface ViewPersonalDeckPageProps {
 export const ViewPersonalDeckPage = ({ deck }: ViewPersonalDeckPageProps) => {
   const navigate = useNavigate();
   const tags = [deck.metaData.access.toLocaleLowerCase(), ...getDeckTags(deck)];
+  const percentStudied = Math.round(deck.metaData.studiedPercent);
```
jiejasonliu, line 21: here as well!

jiejasonliu (commented):
totally fine on just having study percentages, we could work on bringing in mastery if we end up having additional time at the end!

src/pages/study-page/study-results-page/study-results-page.tsx
derricklee43: Oh, I thought we were showing the percentage change before and after the lesson, so the delta, not the total. But I'm fine with doing the total; it makes more sense with the current label

jiejasonliu (approved):

PR #112, opened 2022-10-28
API calls made to be lowercase

masonitea:


jiejasonliu:
this is great! does case-sensitivity matter for `/Refresh` as well?
if so, you may need to do the same for `use-fetch-wrapper:164`

otherwise, LGTM!

masonitea:
I believe I changed that line. Prior to this though /Refresh and /Authenticate were being called when the endpoints were /refresh and /authenticate already. IDK if it would without these changes with the backend being changed but this is just for the sake of consistency.
joebiden

jiejasonliu:
oh wth i'm blind!!! 👀 
but yeah, maybe it's not case-sensitive, but for consistency sake this is definitely better!

jiejasonliu (approved):

derricklee43 (approved):
lgtm!

PR #114, opened 2022-11-10
fix: keep lesson card outcomes on skip or replay

jiejasonliu:
#### lesson card
- skipping or replaying cards no longer overwrite the previous outcome
- this works because now, we only apply the outcome style on the back of the card

#### study
- disable study buttons with there are no cards 
- prevent access to lesson if manually navigating to study page on a deck with no cards


derricklee43 (approved):
LGTM!

PR #116, opened 2022-11-11
fix: improve dropdown options styling

jiejasonliu:
### summary of changes
- fix bug with ellipsis overflow selector on parent (options container)
- add ellipsis overflow to child (single option) as well
- wrap all options in a grid component that evenly distributes height

---

**(search box)**
image

**(card options)**
image

### DIFF
--- a/src/components/drop-down-options/drop-down-options.scss
+++ b/src/components/drop-down-options/drop-down-options.scss
@@ -15,8 +15,15 @@
     min-width: fit-content;
   }
 
-  .ellipsis-overflow {
+  &.ellipsis-overflow {
     @include ellipsis-overflow;
+  }
+
+  .c-options-grid {
+    display: grid;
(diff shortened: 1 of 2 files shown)


src/components/drop-down-options/drop-down-options.scss
```diff
@@ -15,10 +15,17 @@
     min-width: fit-content;
   }
 
-  .ellipsis-overflow {
+  &.ellipsis-overflow {
     @include ellipsis-overflow;
   }
 
+  .c-options-grid {
+    display: inline-grid;
```
jiejasonliu, line 23: actually this should probably have been `display: grid` since we don't need it inline, but it won't make a difference since we don't actually (and will likely never) have text around this grid

derricklee43:
Those look really slick!

PR #117, opened 2022-11-12
Add swap all button

derricklee43:
- Add swap all terms and definitions button
- Created DraftDeck which is a regular deck with DraftCard
  - I don't know how I feel about the draft vs regular cards and decks. Starting to think we should just add these optional fields to the Card interface 

https://user-images.githubusercontent.com/39753553/203844076-f190d819-a4e5-4a8f-ad52-bc2b6ead9caa.mov


jiejasonliu (approved):
i think `DraftCard` was a good idea, but `DraftDeck` by itself doesn't mean a lot or add too much benefit
that said, i'm not strongly opinionated toward either keeping it or not since it's very minor

everything else looks great, good cleanup and feature! 😄

derricklee43:
Ok, I'll leave it in. If it causes any headaches in the next couple of weeks, we can always remove it.

PR #119, opened 2022-11-14
fix: await cards promise before passing as payload

jiejasonliu:
## rebase on main before merging

- fixes cards not being added when a new deck is created
- enforce fetch wrapper to only allow non-promises as payload to avoid future oopsies


derricklee43:
ooof, that is really nice! LGTM!

derricklee43 (approved):

PR #123, opened 2022-12-01
feat: overall flow improvements

jiejasonliu:
- prevent navigation to landing page if logged in (hate to do it, but it makes sense)
- flashcard decks now has a loading indicator bc sometimes it takes forever
- fix infinite loading if `finish lesson` 
- `study again` also saves lesson results now


derricklee43:
🚀looks good! I agree, not seeing the landing page when signed in is sad, but it flows better.

PR #124, opened 2022-12-01
feat: add levenshtein string distance to speech grading

jiejasonliu:
### summary
- adds `fast-levenshtein` package to grading which implements https://en.wikipedia.org/wiki/Levenshtein_distance
- also makes an attempt to normalize card text (only keep language letters/characters and numbers)
  - (e.g.) `mercury(I)` becomes `mercuryI` and `dog [puppy]` becomes `dog puppy`

### grading calculation
- define a base threshold `answer.length - 1` or `2` whichever is lowest
- match across the spoken text with a sliding window which is size of num words in answer
- if any matched 'chunks' is within the threshold, it is considered correct

### examples [spoken, expected-answer]
- `["cat", "bat"]`
  - correct because the distance is 1 <= 2 (threshold)
- `["i think the answer is three cats wait i mean to bats", "two cats"]`
  - correct because the sliding window (N=2) goes to `to bats` which has a distance of 2 <= 2 (threshold)
- `["猫は", "猫"]`
  - incorrect because the distance is 1, but threshold is 0 since `answer.length - 1 = 0`


src/helpers/string.ts
```diff
@@ -20,3 +20,14 @@ export function stringToBoolean(subject: string | undefined, fallback?: boolean)
       throw new Error(`stringToBoolean tried to convert non-boolean: ${subject}`);
   }
 }
+
+/**
+ * @returns a string only with alphanumeric symbols kept (works for most languages)
+ */
+export function replaceNonAlphanumeric(text: string) {
```
derricklee43, line 27: That's pretty smart! I didn't even think about numbers, haha.

derricklee43 (approved):
lgtm!

PR #125, opened 2022-12-05
feat: add official favicon

jiejasonliu:
better than the react default favicon 😂 
lmk if we had a different icon we wanted to use

image


derricklee43 (approved):
perfect! LGTM!

PR #126, opened 2022-12-09
feat: add card limit and remove assertions

jiejasonliu:
- card limit of 500
- remove assertions, it breaks deletions that have not been serialized yet
  - better let it fail on the server side (if it does), than for us to fail it when it really doesn't break anything


u1068393 (approved):
LGTM!

